PR #7481, opened 2015-01-13
New 'detailed' auth plugin flag 'follow_redirects'

kravietz:
This flag enables the 'detailed' plugin to follow HTTP redirects in multi-stage authentication schemes.


w3af/plugins/auth/detailed.py
```diff
@@ -138,6 +149,7 @@ def get_options(self):
              '    - %p for the password parameter name value\n'
              '    - %P for the password value\n'),
             ('method', self.method, 'string', 'The HTTP method to use'),
+            ('follow_redirects', self.follow_redirects, 'string', 'Follow HTTP redirects on multi-stage authentication pages'),
```
andresriancho: I believe that it should be boolean instead of `'string'`

w3af/plugins/auth/detailed.py
```diff
@@ -58,7 +58,18 @@ def login(self):
 
         try:
             functor = getattr(self._uri_opener, self.method)
-            functor(self.auth_url, data)
+
+            while True:
```
andresriancho: If the web application has a redirect loop, then w3af will be here for ever, which is really bad.

w3af/plugins/auth/detailed.py
```diff
@@ -58,7 +58,18 @@ def login(self):
...
+                # actually file the auth HTTP reuqest
+                r = functor(self.auth_url, data)
+                # follow redirects if the feature is enabled
+                if self.follow_redirects and r.get_code() % 300 < 10:
+                    self.auth_url.set_path(r.get_headers().iget('location')[0])
```
andresriancho: There are problems with this approach:
 * Changing the auth_url, which is a user configured setting might bring issues in the future
 * Only "Location" redirects are supported, when there are many other types

Redirecting is difficult, there is a handler for it in w3af (which inherits from the urllib2 handler) https://github.com/andresriancho/w3af/blob/master/w3af/core/data/url/handlers/redirect.py

This code also follows redirects outside the initial application URL, which is usually something we don't want, but I understand where you're coming from with this.

Finally, the next "GET" to the redirect target is only done after some considerable time happens. Thus the initial call to the login() method will never work.

If you've got some minutes I would love to talk to you about this [on IRC](http://w3af.org/community) so we can find a way to make this feature available to all with the best practices

kravietz: Sure, I can join IRC - will be there around 6pm GMT.

w3af/plugins/crawl/spider_man.py
```diff
@@ -185,8 +185,15 @@ def do_ALL(self):
...
+        # convert relative URL to absolute if request came from CONNECT
+        if hasattr(self.server, 'chainedHandler'):
+            base_path = "https://" + self.server.chainedHandler.path
+            path = base_path + self.path
+        else:
+            path = self.path
+
         # Convert to url_object
-        path = URL(self.path)
+        path = URL(path)
```
andresriancho: This fix doesn't seem to be related with the follow_redirects, could you explain what it fixes?

w3af/plugins/grep/strange_headers.py
```diff
@@ -74,7 +79,7 @@ class strange_headers(GrepPlugin):
         "X-CACHE",
         "X-UA-COMPATIBLE",
         "X-PAD",
-        "X-XSS-Protection"]
+        "X-XSS-PROTECTION"]
     )
 
     def __init__(self):
```
andresriancho: Manually merged this file changes into develop

w3af/plugins/auth/detailed.py
```diff
@@ -58,7 +60,24 @@ def login(self):
...
+
+                # follow redirects if the feature is enabled
+                if self.follow_redirects and r.get_code() in GET_HEAD_CODES:
+                    self.auth_url.set_path(r.get_headers().iget('location')[0])
+                    om.out.debug('Auth redirected to new URL', self.auth_url)
+                    self.method = 'GET'
+                    data = ''
+                    redir_count += 1
+                else:
+                    break
```
andresriancho: This is a big improvement from the past PR, thanks for fixing some of the issues that I pointed out (such as the loop). These issues are still pending:
 * You're modifying the user-configured setting `self.auth_url` and `self.method` which is not desirable
 * You're only supporting "location" header instead of all the ones which the framework knows about (used in `web_spider.py:_headers_url_generator`). Maybe you could re-factor the code used in the web_spider.py into a `core` function that both the web_spider and detailed plugin could use? Something like "extract_redirects(http_response)" which returns a list of all the URL instances which the browser would be redirected to. I know that in 99% of the cases this would be a list with only one URL, but we never know...

w3af/plugins/crawl/spider_man.py
kravietz: See my comments at #7484 It's completely separate issue -  I didn't realize the fix will get mixed together with the others, next time I'd probably should create a branch.

w3af/plugins/auth/detailed.py
kravietz: After looking at _web_spider_ I get your point and will try to think about how to implement this in more reusable way.

andresriancho:
@kravietz do you plan on finishing this PR? It seems to be a good feature for users!

kravietz:
After looking at `web_spider.py:_headers_url_generator` I don't think this is the right way to go. The three HTTP headers listed in _URL_HEADERS_ are useful for fuzzing, but in case of the authentication workflow we're following a normal route that a standard browser would follow when logging in. In such case we should just follow the `Location` header.

I have fixed the code not to override the user-supplied `auth_url` though.

P.S. what we do miss here are JavaScript redirects that are common in case of SAML login workflows, but that's a separate feature that goes well beyond the scope of this merge request.

andresriancho:
> After looking at web_spider.py:_headers_url_generator I don't think this is the right way to go. The three HTTP headers listed in URL_HEADERS are useful for fuzzing, but in case of the authentication workflow we're following a normal route that a standard browser would follow when logging in. In such case we should just follow the Location header.

Well, there is a detail here which we need to understand. The great majority of web applications will send 302 redirect with `Location` header, some others (like you mention below) will redirect using some JavaScript code, others will use `<meta redir`, and others (minority) will redirect with a `uri` or `content-location` header in a 302:

```python
URL_HEADERS = {'location', 'uri', 'content-location'}
```

Since a regular browser will understand the `uri` and `content-location` headers, w3af should do too, and we should redirect there. For a great number of use cases w3af needs to do exactly what a browser would, and in this case it's redirecting for location, uri, content-location, `<meta redir` and JavaScript. The more we cover the better.

> P.S. what we do miss here are JavaScript redirects that are common in case of SAML login workflows, but that's a separate feature that goes well beyond the scope of this merge request.

Agreed, we need better JS support, but that's another discussion: Javascript crawler #1796

kravietz:
Hm, are you sure about `Uri` and `Content-Location`? 

I've never seen the any live webapp returning `Uri` and the purpose of `Content-Location` is somewhat different - it's not used to suggest a **new location** but rather  indicate a real location of the resource just returned. For example IIS typically returns the following when your request the website's main page:

    Content-Location: http://121.10.133.169/iisstart.htm

This essentially means that the content of the main page (/) was really served from `/iisstart.htm` but this doesn't mean the browser should go and request that file, because it will get the same content.

'Meta redir` is indeed a valid and used method but it requires digging into the HTML code unfortunately...

andresriancho:
> Hm, are you sure about Uri and Content-Location?

Today? No. When I added the feature? Yes.

> I've never seen the any live webapp returning Uri

I've never seen God, that doesn't mean it doesn't exist ;) This is part of python's urllib2 HTTPRedirectHandler:

```python
    # Implementation note: To avoid the server sending us into an
    # infinite loop, the request object needs to track what URLs we
    # have already seen.  Do this by adding a handler-specific
    # attribute to the Request object.
    def http_error_302(self, req, fp, code, msg, headers):
        # Some servers (incorrectly) return multiple Location headers
        # (so probably same goes for URI).  Use first header.
        if 'location' in headers:
            newurl = headers.getheaders('location')[0]
        elif 'uri' in headers:
            newurl = headers.getheaders('uri')[0]
        else:
            return
```

So `location` and `uri` are covered there.

> the purpose of Content-Location is somewhat different - it's not used to suggest a new location but rather indicate a real location of the resource just returned

You do have a point here, [the rfc](http://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html) (search for content-location) doesn't mention anything about redirects.

So, it seems that for redirects **in this plugin** we need location+uri (at least until we add meta, js, etc.).

For the web_spider, and since we don't use the `URL_HEADERS` in a way that's bound to redirects, the `content-location` should stay as a way to extract new URLs from http response headers.

andresriancho:
@kravietz gets extra brownie points for making me read the RFC :+1:

andresriancho:
@kravietz your code was cleaned and added to the framework at https://github.com/andresriancho/w3af/commit/58cc7b032202d2c39d9385595379dd829677915c , I'm running unittests at https://circleci.com/gh/andresriancho/w3af/1566 and when those are done I'm going to merge it into develop

After a few days that code will make it to master and everyone will be able to benefit from it :+1: 

Thanks for the code, idea, etc.!

Issue #5377, opened 2014-09-30
Cookie plugin secure/httponly improvements


andresriancho:
## User story
As a user I'm seeing vulnerabilities like: "A cookie without the HttpOnly flag was sent when  requesting "https://.../reg/a.html". The HttpOnly flag prevents potential intruders from accessing the cookie value through Cross-Site Scripting attacks. This vulnerability was found in the request with id 2501."

I would like to see the following improvements:
 - [x] The cookie name is shown in the description
 - [ ] Vulnerabilities are reported once per cookie
 - [x] A missing secure flag is not a high risk vulnerability

## Conditions of satisfaction
 - [ ] All user issues are fixed
 - [ ] New unittests are written to test these new features

andresriancho:
Related with Incorrect grouping in analyze_cookies.py #6642

andresriancho:
Related with cross_domain_js: One vulnerability for each page that points to third party script #3955

andresriancho:
@onizenso : While working on a similar issue, but for the cross_domain_js plugin (see #3955) I created a new way of handling this: `InfoSet`.

You can see that in action [here](https://github.com/andresriancho/w3af/blob/feature/multi_urls_3955/w3af/plugins/grep/cross_domain_js.py#L124):

```python
            ff = lambda iset, info: iset.get_attribute(itag) == info[itag]
            self.kb_append_uniq_group(self, 'cross_domain_js', i, ff)
```

Those two lines are full of magic :+1: and allow us to group Info instances based on a tag that's applied to it.

If there is no `InfoSet` already created which matches the `itag`, then a new one is created to hold the info (`i`) which is passed as parameter. If there is a match then `InfoSet.add(i)` is called and the info set instance is updated in the knowledge based.

All this was required in order to have cleaner plugin code and object structure.

Your code for the knowledge base update, and the other fixes from #7763 were merged into develop and already made it to all our users in the latest release, congrats! The code for the plugin itself won't make it to the develop nor master branches, since I'm going to rewrite that using InfoSet objects.

You can consider your work for #5377 a great success, I do! And if you want please continue with https://github.com/andresriancho/w3af/issues/8523

andresriancho:
TODOs for me to complete this task:
 - [ ] Create a new feature branch
 - [ ] Move all unittests from the pull request into the new branch
 - [ ] Re-use @onizenso's code for grouping info instances into InfoSet
 - [ ] Make sure all tests PASS

[...]

PR #7763, opened 2015-01-25
Feature/httponly secure cookie

onizenso:
For the modified tests, I needed to include the test for the cookie name to keep them from breaking. They already included the substring of the description where I added the cookie name, so no need for a new test.

Modified tests:
test_analyze_cookies_secure_over_http
test_analyze_cookies_no_httponly

New tests:
test_analyze_cookies_no_secure_over_https_has_cookie_name
test_analyze_cookies_secure_over_http_has_cookie_name

For a cookie sent over HTTPS with no secure flag, I set the severity to low.


andresriancho:
@onizenso not sure why but there is a revert of some of my work in your pull request. If I would merge this into develop the work I've done in https://github.com/onizenso/w3af/commit/68f4e893f6dbf104f9eb51f170b838156e1c140a would be lost. If this was your intention, could you explain why? If not, I believe the best case scenario is for you to create a new PR without that.

w3af/plugins/grep/analyze_cookies.py
```diff
@@ -204,13 +204,16 @@ def _http_only(self, request, response, cookie_obj,
         :return: None
         """
         if not self.HTTPONLY_RE.search(cookie_header_value):
-            
             vuln_severity = severity.MEDIUM if fingerprinted else severity.LOW
-            desc = 'A cookie without the HttpOnly flag was sent when ' \
+            keys = ''
+            for key in cookie_obj.keys(): 
+                keys += key + ' '
+            keys = keys.strip()
```
andresriancho: I don't like the fact that these lines are repeated all over the code:
```patch
+            keys = ''
+            for key in cookie_obj.keys(): 
+                keys += key + ' '
+            keys = keys.strip()
```

Maybe create a small helper method that does this?

w3af/plugins/grep/analyze_cookies.py
```diff
@@ -204,13 +204,16 @@ def _http_only(self, request, response, cookie_obj,
...
+            desc = 'Cookie(s) (%s) without the HttpOnly flag was sent when ' \
```
andresriancho: I believe that here we have a deep problem that was present in the plugin before you started to work on it, but we just found out about it when trying to implement #5377.

In most cases the application will send only one cookie. In that case we'll find that the message works well:

`Cookie(s) (foobar) without HttpOnly...`

The only problem there is that the message says "Cookie(s)" and the "(s)" is not really needed because there is only one cookie. But then, the real problem occurs... what happens if in one HTTP response we receive two cookies, one with httponly and one without? I believe the code will show:

`Cookie(s) (withouthttponly, withhttponly) without HttpOnly...`

Which is incorrect, since one of the cookies is secure (withhttponly) and we're saying it's not.

Could you please test this?

andresriancho: I believe that we should always have descriptions saying `Cookie "foobar" was sent without httponly", no `(s)` should be present.

w3af/plugins/grep/analyze_cookies.py
```diff
@@ -352,16 +359,22 @@ def _not_secure_over_https(self, request, response, cookie_obj,
...
                    ' channel, thus preventing potential session hijacking' \
                    ' attacks.'
-            desc = desc % response.get_url()
+            desc = desc % (keys,response.get_url())
             
+	    # Severity set to low, secure flag only protects
+	    # confidentiality which is already protected by SSL
             v = Vuln('Secure flag missing in HTTPS cookie', desc,
-                     severity.HIGH, response.id, self.get_name())
+                     severity.LOW, response.id, self.get_name())
```
andresriancho: Yes, I agree with this change.

onizenso:
Apologies for the bad revert, didn't mean to erase your code. I was trying to correct Vim automatically adding an EOL to files that didn't originally have them. I reverted the revert to correct this mistake.

I'll add tests for the other issues, and respond inline to the points you made.

w3af/plugins/grep/analyze_cookies.py
onizenso: I was thinking similarly along those lines, asking myself what if multiple cookies were sent with one HTTP response and only one of them being HttpOnly. I'll create a test for this. 

I'll change the description to match `Cookie "foobar" was sent without httponly"`

w3af/core/data/kb/info.py
```diff
@@ -93,7 +93,7 @@ def from_info(cls, other_info):
             raise TypeError('Info expected in from_info.')
         
         name = other_info.get_name()
-        desc = other_info.get_desc()
+        desc = other_info.get_desc(with_id=False)
```
andresriancho: This seems like an interesting edge case in `from_info` that should get into master!

andresriancho:
@onizenso with the talk we had the other day + some changes this PR would be ready to merge, thanks for your hard work :+1:

w3af/plugins/grep/analyze_cookies.py
```diff
@@ -187,6 +199,27 @@ def _analyze_cookie_security(self, request, response, cookie_obj,
...
+                old_vuln = vulns[0]
+                updated_vuln = copy.deepcopy(old_vuln)
+                if 'urls' not in updated_vuln.keys():
+                    updated_vuln['urls'] = [old_vuln.get_url()]
+                updated_vuln['urls'].append(response.get_url())
+                
+                kb.kb.update(old_vuln, updated_vuln)
+            else:
+                # error
+                raise ex.DBException('An error has occured retrieving' \
```
andresriancho: I believe there is no way to reach this line of code. Remove all the "else" section of this if statement after confirming.

Also remove the `ex` import above.

w3af/plugins/grep/analyze_cookies.py
```diff
@@ -187,6 +199,27 @@ def _analyze_cookie_security(self, request, response, cookie_obj,
...
 
+    def _update_cookie_vulnerability(self, response, vulnerability):
+        # thread lock to escape race condition on db access
+        with self._plugin_lock:
+            vulns = kb.kb.get(self, 'security')
+        
+            if len(vulns) == 0:
+                kb.kb.append(self, 'security', vulnerability)
+            elif len(vulns) >= 1:
+                old_vuln = vulns[0]
```
andresriancho: Hmmm... there might be an issue here. Have you tested what happens when a site has cookies which lack the secure AND httponly flags? What I believe will happen is:
 * Plugin receives a response with a cookie without httponly
 * First call to `_update_cookie_vulnerability` will return zero vulns in `vulns = kb.kb.get(self, 'security')`
 * The httponly vulnerability is stored in the kb
 * Call to `_update_cookie_vulnerability` will return the httponly vulnerability we stored in the previous step
 * The new URL (which is vulnerable to secure flag missing) is added to the first vulnerability, which is actually related to httponly

andresriancho:
Partial merges of your work were done:
 * https://github.com/andresriancho/w3af/commit/d18c84bb4d616daa149ce3fe92d06e32d21128b2
 * https://github.com/andresriancho/w3af/commit/f544be0983c75b569da9ba84027fb188fafc33cb
 * https://github.com/andresriancho/w3af/commit/c0ad09af3e42309c6945ef675ced9780c99cf3d8

These are valuable and can be used by others to build their plugins.

The improvements for the httponly+secure plugin still need some work.

w3af/plugins/grep/analyze_cookies.py
onizenso: You were right on the method updating the vulnerability outside its type. I updated the method to now test on both the amount of vulnerabilities from the KB and the name of each returned vulnerability.

> The new URL (which is vulnerable to secure flag missing) is added to the first vulnerability, which is actually related to httponly

This is now fixed with:
https://github.com/andresriancho/w3af/pull/7763/files#diff-dceefea625bb1de5a6c0438ca6b36c2eR201
```
def _update_cookie_vulnerability(self, response, vulnerability):
 # thread lock to escape race condition on db access
 with self._plugin_lock:
 vulns = kb.kb.get(self, 'security')

 name = vulnerability.get_name()

 if len(vulns) == 0:
 kb.kb.append(self, 'security', vulnerability)
 else:
 for vuln in vulns:
 # Check if vuln and vulnerability are the same type
 if vuln.get_name() == name:
 old_vuln = vuln
 updated_vuln = copy.deepcopy(old_vuln)
 if 'urls' not in updated_vuln.keys():
 updated_vuln['urls'] = [old_vuln.get_url()]
 updated_vuln['urls'].append(response.get_url())
 kb.kb.update(old_vuln, updated_vuln)
 return updated_vuln
 # If no vulnerability of same type was found
 # append the vulnerability passed by caller
 kb.kb.append(self, 'security', vulnerability)
```
and tested with 
https://github.com/andresriancho/w3af/pull/7763/files#diff-36d002ea93080120a5535f35c1ed8417R344
```
def test_update_without_httponly_secure_over_https(self):
 body = ''
 url = URL('https://www.w3af.com/')
 headers = Headers({'content-type': 'text/html',
 'Set-Cookie': 'name="adf"'}.items())

 response = HTTPResponse(200, body, headers, url, url, _id=1)
 request = FuzzableRequest(url, method='GET')

 for i in range(0, 2):
 self.plugin.grep(request, response)

 self.assertEqual(len(kb.kb.get('analyze_cookies', 'cookies')), 1)
 self.assertEqual(len(kb.kb.get('analyze_cookies', 'security')), 2)
```

andresriancho:
See note https://github.com/andresriancho/w3af/issues/5377#issuecomment-76402565 on why I'm closing this PR

PR #7788, opened 2015-01-26
Debian 7.8 Platform Addition

codychamberlain:
I added platform check for Debian 7.8.

Working on adding as many platforms as I can.

Tested on two vanilla DigitalOcean 7.8 Debian images and worked great.

### DIFF
--- a/w3af/core/controllers/dependency_check/platforms/debian78.py
+++ b/w3af/core/controllers/dependency_check/platforms/debian78.py
@@ -1,5 +1,5 @@
 """
-current_platform.py
+debian78.py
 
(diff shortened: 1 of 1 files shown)


andresriancho:
Thanks for the pull request, very much appreciated. I'm very interested in people submitting these since they really help others with their installation process.

One note, for the next PRs you send, please make them against the "develop" branch. In this case I took your code and manually added it to that branch, but if you could send the next ones directly there, that would be awesome :+1: 

https://github.com/andresriancho/w3af/commit/755f06d5d9d1d306cedc82aa5f8c306e5621e1a3

PR #7821, opened 2015-01-27
Added Ubuntu 14.10

codychamberlain:
Based on testing of DigitalOcean image added Ubuntu 14.10 platform. Had to add python-psutil to CORE dependency list as this is apparently no longer installed by default (at least by DigitalOcean)

### DIFF
--- a/w3af/core/controllers/dependency_check/platforms/ubuntu1410.py
+++ b/w3af/core/controllers/dependency_check/platforms/ubuntu1410.py
@@ -1,0 +1,33 @@
+"""
+ubuntu1410.py
+
+Copyright 2014 Andres Riancho
+
+This file is part of w3af, http://w3af.org/ .
+
+w3af is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation version 2 of the License.
+
+w3af is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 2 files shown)


andresriancho:
psutil requirement was a mistake from my side, I've corrected it with https://github.com/andresriancho/w3af/commit/e6914ce01f3f417330edfc6480bfcdbc3edac193

w3af/core/controllers/dependency_check/platforms/ubuntu1410.py
```diff
@@ -0,0 +1,72 @@
...
+
+"""
+import subprocess
+import platform
+
+from .base_platform import Platform
+from ..requirements import CORE, GUI
+
+
+class Ubuntu1410(Platform):
```
andresriancho: Why don't you inherit from `Ubuntu1404`? That would make the `Ubuntu1410` class really small, avoid code duplication at `os_package_is_installed`, etc

andresriancho:
When sending the corrected PR, if possible, re-base it on top of https://github.com/andresriancho/w3af/commit/e6914ce01f3f417330edfc6480bfcdbc3edac193

w3af/core/controllers/dependency_check/platforms/ubuntu1410.py
codychamberlain: i tried inheriting the class originally but when i needed to include psutil i couldn't override the CORE_SYSTEM_PACKAGE list. now that psutil is removed i'll go back to inheriting the ubuntu class. agreed!

PR #7847, opened 2015-01-27
Ubuntu 14.10 Platform

codychamberlain:
Now that psutil was removed a reverted back to my original commit (d5214ef) where I inherited the Ubuntu 12 platform class.

Fetched upstream changes removing psutil and tested on DigitalOcean Ubuntu 14.10 image and worked great.


andresriancho:
:dancer: :dancers: :+1:

PR #7979, opened 2015-02-03
Update current_platform.py

alener:


andresriancho:
This doesn't even "compile", the syntax is incorrect

PR #7988, opened 2015-02-03
Add w3af_batch, multihost console wrapper for w3af

skoval00:
This script may be used to scan multiple targets on different domains in parallel with a given profile. It uses multiprocessing under the hood where each target is scanned by individual process. It also supports scan's timeouts.
I got stuck with KeyboardInterrupt handling and made some dirty solution where scan terminated with several stacktraces if Ctrl+C is pressed.


w3af_batch
```diff
@@ -0,0 +1,134 @@
...
+        if STOPSCAN.wait():
+            pool.terminate()
+
+    logger.info('Begin scanning')
+    pool = mpd.Pool(args.workers)
+    controller = th.Thread(target=terminate_scan,
+                           name='Controller',
+                           args=(pool,))
+    controller.start()
+    with open(args.targets) as targets:
```
andresriancho: With the argparse trick below we don't need error handling here, but if it doesn't exist then we'll have to add error handling (file doesn't exist, permission error, etc.)

w3af_batch
```diff
@@ -0,0 +1,134 @@
...
+
+
+if __name__ == '__main__':
+    signal.signal(signal.SIGINT, cancel_scan)
+    parser = argparse.ArgumentParser()
+    parser.add_argument('-p', dest='workers', type=int,
+                        default=mp.cpu_count(),
+                        help='number of worker processes '
+                             '(default: number of CPUs)')
+    parser.add_argument('-t', default=300, dest='timeout', type=int,
```
andresriancho: The help needs to specify if this is seconds, minutes, etc.

w3af_batch
```diff
@@ -0,0 +1,134 @@
...
+        timer.cancel()
+        timer.join()
+        logger.debug('Target %s has %s vulnerabilities',
+                     target, len(kb.get_all_vulns()))
+
+
+def start_executor(target, profile, timeout):
+    """
+    Start w3af scan in a separate process. Terminate child process after
+    2 * timeout seconds.
```
andresriancho: Why after 2*timeout? Won't this confuse users?

w3af_batch
```diff
@@ -0,0 +1,134 @@
...
+
+    def terminate_scan(pool):
+        """
+        Terminate pool tasks processing in case of SIGINT.
+        """
+        if STOPSCAN.wait():
+            pool.terminate()
+
+    logger.info('Begin scanning')
+    pool = mpd.Pool(args.workers)
```
andresriancho: Any special reason for this being a threadpool instead of a process pool? If you use a process pool you might be able to simplify your code (remove the mp.Process above)

w3af_batch
```diff
@@ -0,0 +1,134 @@
...
+    """
+    Dummy SIGINT hanlder which stops target list processing.
+    """
+    STOPSCAN.set()
+    raise KeyboardInterrupt('Scan is terminated by user request')
+
+
+def execute_scan(target, profile, timeout):
+    """
+    Execute w4af scan.  Try to stop it after timeout seconds.
```
andresriancho: w3af

andresriancho:
Thanks for the PR, I like your code! Please fix the issues and after that I'll do some manual testing

w3af_batch
skoval00: It should act as a precaution. I want to stop long-running scans with w3afCore.stop(). Maybe it would be better to add some additional timeout and then terminate worker process. What do you think?

w3af_batch
skoval00: My coworker asked me to run each scan in a separate process. He wants them to be completely isolated. But I suddenly realized that I could create new w3afCore object for each target/task run to achieve isolation.
So there is only one good reason left. I don't know how to terminate long-running scans in a pool from the main process. I need to check how pool behaves if some of its workers stop execution.

w3af_batch
andresriancho: I believe that users expect timeouts to work :) So, what I would do is:
 * Call w3afCore.stop() when the configured timeout is met
 * Send a log message saying something like: "Stopping scan ... because of configured timeout ..."
 * Let w3afCore.stop() try to stop the scan for 30 seconds. If that doesn't work terminate() the process and log a message like: "Terminated process after waiting for w3afCore.stop()". If the stop did work log some other debug message.

I leave the logging level of each of those messages up to you

w3af_batch
andresriancho: Understood, I believe it's a good idea to keep the threadpool+process spawning then.

skoval00:
Andres, thank you for your review and suggestions. Seems like I've fixed everything that you mentioned.
There is one problem left that disturbs me. This is handling of Ctrl+C keyboard interrupt. I'll probably look at it a bit later. Right now scanner eventually stops, sometimes with a delay. I suppose after waiting for some threads.

w3af_batch
```diff
@@ -0,0 +1,136 @@
...
+STOPSCAN = mp.Event()
+logger = logging.getLogger('w3af_batch')
+
+
+def cancel_scan(*args):
+    """
+    Dummy SIGINT hanlder which stops target list processing.
+    """
+    STOPSCAN.set()
+    raise KeyboardInterrupt('Scan is terminated by user request')
```
andresriancho: Any special reason you raise an exception in the signal handler?

w3af_batch
```diff
@@ -0,0 +1,136 @@
...
+    scanner = w3afCore()
+    scanner.profiles.use_profile(profile)
+    option_list = scanner.target.get_options()
+    option_list['target'].set_value(target)
+    scanner.target.set_options(option_list)
+    scanner.plugins.init_plugins()
+
+    timer = th.Timer(timeout, stop_scan, (scanner,))
+    timer.start()
+    scanner.start()
```
andresriancho: The w3afCore might raise exceptions, please handle them. You might find which exceptions are raised in the console UI / GUI code.

w3af_batch
```diff
@@ -0,0 +1,136 @@
...
+    thread executes scan in a separate process. Scanner process will
+    try to stop w3af after timeout seconds. In addition manager thread
+    will terminate the child process after WAIT_TIMEOUT seconds.
+    """
+    def terminate_scan(pool):
+        """
+        Terminate processing of pool tasks in case of SIGINT.
+        """
+        if STOPSCAN.wait():
+            pool.terminate()
```
andresriancho: The Ctrl+C handling "timeout" might be hanging in this line, because `pool.terminate()` doesn't kill the pool threads, it just clears the task queue and calls join() on the running threads. To make sure that it all exits quickly you might want to use the STOPSCAN to (also) call `.terminate()` on the running executor(s)

andresriancho:
@skoval00 great improvements from the initial PR! I love all your changes. The comments I made now are mostly related to your concern about Ctrl+C handling, what I'm trying to do with this second round of comments is to give you some tips on where I believe the code might be improved to better handle that signal.

Hopefully you'll figure it out and we'll have a `w3af_batch` in the next release :+1:

oxdef:
By the way coworker is me :)

andresriancho:
@oxdef always you, behind the shadows!

andresriancho:
@skoval00 ping! You going to finish this PR?

skoval00:
Yes, I think I'll be able to return to it at the end of the week. Thank you for a reminder :)

andresriancho:
Thanks. No pressure, but if you hurry it will get into the Kali package I'm building and (depends on Kali devs) would go out to the public next week.

andresriancho:
Ping!

skoval00:
Hello, Andres. In my first attempt to write w3af_batch I struggled to test the script a lot. So I decided to rewrite it and added some tests. It is not finished yet and some things should be polished. And the main obstacle is testing SIGINT handling (test_w3af_batch.PoolTest.test_interrupt_stops_execution). I'm not sure it is possible to test such behaviour correctly. Right now signal handler in test is activated after the mock Job is complete.
But the good news is that w3af_batch run stops immediately after Ctrl+C is pressed.

The workflow remains without changes. There is a Pool of Manager threads, every Manager controls its own Worker.

So I decided to ask your opinion about the last broken test. And about the structure of my commit. I am concerned that to test w3af_batch I need w3af_batch.py as python module, but I don't like that name for a script.

Thank you for your persistence and great work :)

andresriancho:
## Comments

Did some changes on your code, here are the comments:

 * Added dependency check to make sure w3af can be run
 * Moved w3af_batch.py to w3af/core/ui/batch/
 * w3af_batch is now "a pointer" (not in the file system way as before, but an import) to w3af/core/ui/batch/w3af_batch.py , this solves one of the problems you had with the test not finding the file
 * Having the test in the root path was ugly, moved to w3af/core/ui/batch/
 * Added copyright headers
 * Added you as the author of the classes
 * Changed arguments to match w3af_console and w3af_gui (-P to -p)
 * Run my first `w3af_batch` scan and I got tracebacks:
```
Process Process-2:
  File "/usr/lib/python2.7/multiprocessing/process.py", line 258, in _bootstrap
    self.run()
  File "/usr/lib/python2.7/multiprocessing/process.py", line 114, in run
    self._target(*self._args, **self._kwargs)
  File "/home/pablo/pch/w3af-merge-helper/w3af/core/ui/batch/main.py", line 82, in run
    worker_job.start()
  File "/home/pablo/pch/w3af-merge-helper/w3af/core/ui/batch/main.py", line 188, in start
    self.scanner.start()
  File "/home/pablo/pch/w3af-merge-helper/w3af/core/controllers/w3afCore.py", line 235, in start
    self.scan_end_hook()
  File "/home/pablo/pch/w3af-merge-helper/w3af/core/controllers/w3afCore.py", line 442, in scan_end_hook
    parser_cache.dpc.stop_workers()
AttributeError: 'module' object has no attribute 'dpc'
```
I believe that these errors come from:
```python
if is_main_process():
    manager = Manager()
    dpc = ParserCache()
```
Did you try w3af_batch with the latest w3af from develop?

## TODOs:
 * We need to find out a good way to provide logging for users in this case. Let's say I scan site A and B with the same profile. The profile uses foo.txt as text_file output. Both scans will try to write to foo.txt and things will break. There should be a way to tell w3af_batch to save the output to different files. Maybe the user could provide a profile **template** instead of a profile; and we can render the template before passing it to w3af? This way the user could use some predefined variables such as `{{ target_domain }}` and `{{ scan_id }}` (integer, increments on each new scan) and have different output files? If you go this path please use `jinja2` since we already have that as a requirement for w3af

 * Add documentation for new w3af_batch command to `doc/sphinx/`

 * In the documentation add an example file for:
```python
    parser.add_argument('-l', default=None, dest='logconfig',
                        help='Config file for Python\'s logging module')
```
 * `test_interrupt_stops_execution` fails on my workstation:
```
[pablo:/home/pablo/pch/w3af-merge-helper] [w3af] skoval00-w3af_batch 3s ± nosetests w3af/core/ui/batch/tests/
..F....
======================================================================
FAIL: test_interrupt_stops_execution (w3af.core.ui.batch.tests.test_w3af_batch.PoolTest)
----------------------------------------------------------------------
  File "/home/pablo/pch/w3af-merge-helper/w3af/core/ui/batch/tests/test_w3af_batch.py", line 186, in test_interrupt_stops_execution
    self.assertAlmostEqual(1, time() - start)
  File "/home/pablo/pch/w3af-merge-helper/w3af/core/ui/batch/tests/test_w3af_batch.py", line 100, in assertAlmostEqual
    super(BaseTest, self).assertAlmostEqual(first, second, delta=0.3)
AssertionError: 1 != 2.129723072052002 within 0.3 delta

----------------------------------------------------------------------
Ran 7 tests in 6.006s

FAILED (failures=1)
```

## Branch
 * I've been working on https://github.com/andresriancho/w3af/commits/skoval00-w3af_batch , I recommend you merge that branch into your `w3af_batch` branch and complete the TODOs (and any other work you can think of) there.

skoval00:
Hi, Andres. Thank you for your thorough response. I'll try to look at it soon.

skoval00:
Hi, Andres. I started processing your TODO list and got a few questions.
1. I set name='MainProcess' for Worker processes. Because otherwise they are unable to import dpc from w3af.core.data.parsers.parser_cache. Is it fine?
2. I merged your changes and rebased my branch on develop. Should I continue pushing to this branch/PR?

skoval00:
Btw, I also encountered two issues.
1) Exception in PoolResultHandler http://www.pastebin.ca/3017663
2) w3af_console -P test.pw3af command does not stop on Ctrl+C in reasonable time. w3af prints "User pressed Ctrl+C, stopping scan." and then nothing happens for a long time. I had no patience to wait and killed 2 processes after 10 minutes.
It only happens on one old site with lots of pages. Is there something we should do about it?

andresriancho:
> 1. I set name='MainProcess' for Worker processes. Because otherwise they are unable to import dpc from w3af.core.data.parsers.parser_cache. Is it fine?

That's kind of cheating right? Not sure what the right solution for this is, but if you think it's the best, I'll believe it.

> 2. I merged your changes and rebased my branch on develop. Should I continue pushing to this branch/PR?

Yes

andresriancho:
> 1) Exception in PoolResultHandler http://www.pastebin.ca/3017663

Looks like an unrelated bug, created an issue here:
https://github.com/andresriancho/w3af/issues/10267

Could you please add more information for me to debug/reproduce/fix?

> 2) w3af_console -P test.pw3af command does not stop on Ctrl+C in reasonable time. w3af prints "User pressed Ctrl+C, stopping scan." and then nothing happens for a long time. I had no patience to wait and killed 2 processes after 10 minutes. It only happens on one old site with lots of pages. Is there something we should do about it?

Well, this might be a bug also. It seems that w3af is waiting for all threads to finish and that's taking too much time. If you find a reliable way to reproduce please create a new issue with all the info and I'll debug/fix

andresriancho:
Fixed https://github.com/andresriancho/w3af/issues/10267 , forget about that one

andresriancho:
Ping!

andresriancho:
@skoval00 ping! Are you going to finish this PR?

skoval00:
Hi! I'll try to finish it in a few days. Thank you for being persistent :)

andresriancho:
Doh! You removed the branch? Creating a new one to complete the task?

skoval00:
Oh! I missed something and probably removed the wrong branch. There was a problem with merging.
I am still trying to figure out how to test SIGINT handling.

PR #8037, opened 2015-02-04
linuxmint17 added

alener:


andresriancho:
PR is still against w3af `master` branch, it should be against `develop`. [Here](https://github.com/andresriancho/w3af/wiki/Contributing-101) is some documentation on how to do that, specifically the that says "In the following screenshot you'll see that the target branch is master, you need to change that to develop by clicking on "Edit":"

w3af/core/controllers/dependency_check/platforms/current_platform.py
```diff
@@ -30,9 +30,10 @@
 from .openbsd import OpenBSD5
 from .suse import SuSE
 from .default import DefaultPlatform
+from .linuxMint17 import LinuxMint17
```
andresriancho: File name should be either:
 * All lower case `linuxmint17`
 * Underscore separated `linux_mint_17`

camelCase is not accepted

PR #9233, opened 2015-03-31
HPP plugin with unittests

nixwizard:


w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+LAST = 'last occurance'
+FIRST = 'first occurance'
+CONCAT = 'concatination'
+UNK = 'unknown'
+
+
+class hpp(AuditPlugin):
+    """
+    Identify http parameter pollution vulnerabilities.
+    """
```
andresriancho: Add yourself as author here

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def __init__(self):
+        AuditPlugin.__init__(self)
+        self.parser = etree.HTMLParser(remove_blank_text=True, remove_comments=True)
+        self.precedence = None
+        self.rlink = ''
+
+
```
andresriancho: Remove extra empty line between methods (pep-8)

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                                v[-len(pattern2):] == pattern2:
+                    result = p, ref[0]
+
+        return result
+
+
+    def strip_trash(self, doc):
+        #stripping HTML comments
+
+        tree   = etree.parse(StringIO(doc), self.parser)
```
andresriancho: Remove extra spaces between tree and "="

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                    result = p
+                #or for concatenation of both were
+                if v[:len(pattern1)] == pattern1 and \
+                                v[-len(pattern2):] == pattern2:
+                    result = p, ref[0]
+
+        return result
+
+
+    def strip_trash(self, doc):
```
andresriancho: Why are comments and extra spaces so bad? Is the performance hit of parsing HTML worth?

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+        self.precedence = None
+        self.rlink = ''
+
+
+    def audit(self, freq, orig_response):
+        """
+        Tests an URL for HPP vulnerabilities.
+
+        :param freq: A FuzzableRequest
+
```
andresriancho: Remove extra empty space in method comment

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+        rparams = {}
+
+        if not orig_response.is_text_or_html():
+            return
+
+        if not orig_response.get_code() == 200:
+            return
+        method = freq.get_method()
+        self._orig_uri = freq.get_uri()
```
andresriancho: `self._orig_uri = freq.get_uri()` won't work, since we run audit plugins in different threads and thread #1 will overwrite the `self._orig_uri = freq.get_uri()` set by thread #2 . You should use a local variable

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+class hpp(AuditPlugin):
+    """
+    Identify http parameter pollution vulnerabilities.
+    """
+
+    def __init__(self):
+        AuditPlugin.__init__(self)
+        self.parser = etree.HTMLParser(remove_blank_text=True, remove_comments=True)
+        self.precedence = None
+        self.rlink = ''
```
andresriancho: Remove unused attr

w3af/plugins/audit/hpp.py
nixwizard: Don't you think there may be some dynamic content's in comments?

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                continue
+            rp = self.find_reflection( dp, REFLECT_TOKEN)
+
+            if not rp:
+                # no reflection at all skip the parameter
+                continue
+            rparams[param] = rp
+            # print "querystring parameter %s is reflected inside " \
+            #       "a link HTML code" % param
+            if self.precedence:
```
andresriancho: Please make sure that you're taking threads into account when working with self.precedence, there might be a race condition here too similar to the one with `self._orig_uri`

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+            mutants = create_mutants(freq, [REFLECT_TOKEN],
+                                     fuzzable_param_list= [param],
+                                     mutant_tuple=(QSMutant,))
+            mu = mutants[0]
+            #first we send REFLECT_TOKEN
+            resp = self._uri_opener.send_mutant(mu)
+            try:
+                dp = parser_cache.dpc.get_document_parser_for(resp)
+            except BaseFrameworkException:
+                return
```
andresriancho: Not 100% sure, but maybe here you should use `continue` instead of `return`?

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+            #do we have a reflection?
+
+            body = self.strip_trash(resp.body)
+            if not resp.get_code() == 200:
```
andresriancho: Since checking the 200 code is VERY fast, and doing the parser stuff is rather slow, I would move the the `if not resp.get_code() == 200:` check right after the `send_mutant`

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                #no need to seek reflection if it is not OK
+                continue
+            rp = self.find_reflection( dp, REFLECT_TOKEN)
```
andresriancho: Remove extra space after (

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+
+    def audit(self, freq, orig_response):
+        """
+        Tests an URL for HPP vulnerabilities.
+
+        :param freq: A FuzzableRequest
+
+        """
+
+        rparams = {}
```
andresriancho: Only define it if you'll use it, I would define it right before the for loop

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+            if not rp:
+                # no reflection at all skip the parameter
+                continue
+            rparams[param] = rp
+            # print "querystring parameter %s is reflected inside " \
+            #       "a link HTML code" % param
+            if self.precedence:
+               continue
+
+            #this hack is to make w3af send request with unencoded qs
```
andresriancho: Do we need this hack? Doesn't it work if you do something like:

```
qs[param].append(REFLECT_TOKEN)
qs[param].append(REFLECT_TOKEN)
```

That way when you do str(qs) you get something like: `param=REFLECT_TOKEN&param=REFLECT_TOKEN` ?

w3af/plugins/audit/hpp.py
andresriancho: So, if at any point in time we found/set a precedence then nothing else is done. There is something wrong with this logic, since:

 * Call #1 to audit() sends N requests and sets precedence
 * Call #2 to audit() sends M requests and no matter what they return, nothing is done.
 * Call #N to audit() sends Y requests and no matter what they return, nothing is done.

So we're wasting a lot of HTTP requests once the precedence is already set. I would put something like this at the top of the audit method code:

```
if self.precedence:
    return
```

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                continue
+
+            ptest_body = self.strip_trash(ptest_resp.body)
+
+            try:
+                ptest_dp = parser_cache.dpc.get_document_parser_for(ptest_resp)
+            except BaseFrameworkException:
+                return
+
+            if fuzzy_equal(self.strip_trash(orig_response.body),
```
andresriancho: `self.strip_trash(orig_response.body)` which is a CPU intensive task is calculated once for each QS parameter, move this outside the for loop

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                           ptest_body, EQUALITY_THRESHOLD):
+                #confirm vuln
+                if self.find_reflection( ptest_dp, value) == rp:
```
andresriancho: Remove extra space after (

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+            if fuzzy_equal(body, ptest_body, EQUALITY_THRESHOLD):
+                #confirm vuln
+                if self.find_reflection( ptest_dp, REFLECT_TOKEN) == rp:
+                    self.precedence = LAST
+                    continue
+                #  report vulnerability
+            try:
+                reflection, link = self.find_reflection( ptest_dp, value, REFLECT_TOKEN)
+            except TypeError:
+                self.precedence == UNK
```
andresriancho: `continue` missing after this line?

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+            if reflection == rp:
+                self.precedence = CONCAT
+                desc = u'Concatenation of values of the two input parameters ' \
+                       u'with the same name \"%s\" detected inside HTML ' \
+                       u'code of response body.\nThe request url was %s.\n' \
+                       u'Got concatenated output reflected in this link %s\n' \
+                       u'This \"feature\" may be used in different WAF ' \
+                       u'bypassing techniques.'
+                desc = desc % (param, ptest_url, link)
+                print desc
```
andresriancho: Remove debug print

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                self.kb_append(self, 'hpp', i)
+                return
+
+
+
+        if self.precedence == UNK:
+            om.out.debug(u'Unable to determine precedence for HPP detection.'
+                         u'Presumably not vulnerable to HPP.')
+
+            return
```
andresriancho: Maybe this return should be a continue? Maybe the precedence was not possible to "calculate" with this QS parameter but will be with the next? Not sure about this... just asking.

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+#        print "Attack!!", self.precedence
+        om.out.debug(u"Detected precedence is %s."
```
andresriancho: I would move all the code after this line to a different method, so it's easier to read/understand/test

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+        vtest_mutants = create_mutants(freq, payloads,
+                                       fuzzable_param_list=rparams.keys(),
+                                       mutant_tuple=(QSMutant,))
+        for m in vtest_mutants:
+            vtest_resp = self._uri_opener.send_mutant(m)
+            try:
+                vtest_dp = parser_cache.dpc.get_document_parser_for(vtest_resp)
+            except BaseFrameworkException:
+                return
+            r =  self.find_reflection( vtest_dp, FORGE_TOKEN)
```
andresriancho: Remove extra space after (

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+            if r:
+                vuln_params.append(r)
+                msg=u"value of parameter %s can be forged with " \
```
andresriancho: Start message with V

andresriancho: Spaces around =

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                    u"HPP attack using URL: %s, precedence is %s"
+                msg = msg % (r, m.get_uri(), self.precedence)
+                v=Vuln.from_mutant(u'HPP', desc=msg, mutant = m,
```
andresriancho: Spaces around =

I recommend using pycharm, which will highlight the pep-8 incorrect code

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+        # strings in web page links or input parameters
+
+        result = False
+        #find reflection in links
+        for ref in parser.get_references():
+            if ref == []:
+                continue
+            qs = ref[0].get_querystring()
+            for p in qs:
+                if self.precedence == LAST and len(qs[p]) == 2:
```
andresriancho: self.precedence race condition

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+    def find_reflection(self, parser, pattern1, pattern2 = ''):
+        #finds reflection of a string or two concatenated
+        # strings in web page links or input parameters
+
+        result = False
+        #find reflection in links
+        for ref in parser.get_references():
+            if ref == []:
+                continue
+            qs = ref[0].get_querystring()
```
andresriancho: Are we sure that when the HTML says `?abc=1&abc=2` the querystring class will keep the order and you'll ALWAYS access the value 1 by doing qs['abc'][0] ?

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+        """
+        return u"""
+        This plugin detects HTTP Parameter Pollution vulnerabilities.
+        """
+
+class HPPInfoSet(InfoSet):
+
+        ITAG = 'hpp_link'
+        TEMPLATE = (
+            'The application is vulnarable to HPP. '
```
andresriancho: Please add a longer description, using the wording from other plugins.

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+            if r:
+                vuln_params.append(r)
+                msg=u"value of parameter %s can be forged with " \
+                    u"HPP attack using URL: %s, precedence is %s"
+                msg = msg % (r, m.get_uri(), self.precedence)
+                v=Vuln.from_mutant(u'HPP', desc=msg, mutant = m,
+                                   plugin_name='hpp',response_ids=[],
+                                   severity=severity.MEDIUM)
+                v.set_url(self._orig_uri)
+                v[HPPInfoSet.ITAG] = m.get_uri()
```
andresriancho: This means that the user will see one vulnerability in the report for each HTTP. The grouping is done by the mutant URI, which will be unique for each vulnerability. It seems that it might not be required to use InfoSets because they are not grouping anything?

w3af/plugins/tests/audit/test_hpp.py
```diff
@@ -0,0 +1,241 @@
...
+              MockResponse(url='http://aspnettarget/test.asp?a=w3afHPPReflectToken&id=1234',
+                           body='<a href="viewemail.asp?client_id=1234&action=w3afHPPReflectToken"> View </a>',
+                           ),
+              ]
+
+    def test_aspnet1(self):
+        cfg = RUN_CONFIG['cfg']
+        self._scan(self.target_url, cfg['plugins'])
+        vulns = self.kb.get('hpp', 'hpp')
+        self.assertEquals(1, len(vulns))
```
andresriancho: In all cases it might be a good idea to also assert the detected precedence

w3af/plugins/tests/audit/test_hpp.py
```diff
@@ -0,0 +1,241 @@
+"""
+test_hpp.py
+
+Copyright 2012 Andres Riancho
```
andresriancho: 2015

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
+"""
+hpp.py
+ 
+Copyright 2006 Andres Riancho
```
andresriancho: 2015

w3af/plugins/audit/hpp.py
andresriancho: Hmmm, there might me dynamic content in HTML comments, true. But the probabilities are slim <5% ? <1% ?

There might also be dynamic content outside of HTML comments affecting the fuzzy_equal function, which has a much greater probability (compared with the previous) and we're not doing anything.

Also, what if the change in the HTML comment is actually related with the change in QS params you sent; and by removing the comments you're not finding it?

I would remove the strip_trash

w3af/plugins/audit/hpp.py
andresriancho: Ah, I believe I better understand this now, and what I said before doesn't make sense. Maybe it should be changed to:

```python
# qs[param][0] before this line has the value REFLECT_TOKEN
qs[param][0] = '%s&%s=%s' % (value, param, REFLECT_TOKEN)
# now it has the desired value to test HPP
```

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                                   plugin_name='hpp',response_ids=[],
+                                   severity=severity.MEDIUM)
+                v.set_url(self._orig_uri)
+                v[HPPInfoSet.ITAG] = m.get_uri()
+                # debug_msg = debug_msg % (rid, freq.get_uri())
+                self.kb_append_uniq_group(self, 'hpp', v,
+                                          group_klass=HPPInfoSet)
+
+    def find_reflection(self, parser, pattern1, pattern2 = ''):
+        #finds reflection of a string or two concatenated
```
andresriancho: Change `#` commend below find_reflection definition to a docstring, so it gets picked up by any tools that create source code documentation

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+                  v =  qs[p][0]
+                #looking for single reflection if only one pattern was entered
+                if v == pattern1 and not pattern2:
+                    result = p
+                #or for concatenation of both were
+                if v[:len(pattern1)] == pattern1 and \
+                                v[-len(pattern2):] == pattern2:
+                    result = p, ref[0]
+
+        return result
```
andresriancho: It's a bad practice to have functions/methods that return different data types.
```python
                    result = p
...
                    result = p, ref[0]

        return result
```
I would recommend, at least, changing the first `result = ` to `result = p, None`. Something similar needs to be done for `result = False`

andresriancho:
# Algorithm analysis
These are the main phases:
 * [Find out if the REFLECT_TOKEN is echoed back in the response](https://github.com/nixwizard/w3af/blob/hpp/w3af/plugins/audit/hpp.py#L83-L104)
 * [Determining QS parameter precedence](https://github.com/nixwizard/w3af/blob/hpp/w3af/plugins/audit/hpp.py#L108-L173)
 * ["Exploitation"](https://github.com/nixwizard/w3af/blob/hpp/w3af/plugins/audit/hpp.py#L176-L207)

# Improvements
## All
Each of the main phases enumerated above should be in a different method. Functions/method with more than 100 lines should be avoided.

## Current HPP detection (Server-side HTTP Parameter Pollution vulnerability)
Right now the server-side HTTP Parameter Pollution vulnerability detection is made by:
 * Sending one HTTP request with REFLECT_TOKEN to check if the parameter is reflected (not sure why this is required?)
 * Sending one HTTP request with `value +'&'+param+'='+REFLECT_TOKEN` to check for precedence

The precedence check is done by analyzing the HTTP responses using fuzzy_equal.

I believe you mixed client and server side HPP, is there a real reason for using find_reflection during server side detection? What about this new algorithm for server side HPP?
 * We receive URL http://host.tld/?id=123 (save as A)
 * Send http://host.tld/?id=122 (save as B)
 * If fuzzy_equal(A, B) then we know that the id parameter is not used to render responses and we can't use it to detect server side HPP
 * Send http://host.tld/?id=123%26id%3D122 (save as C)
 * Compare A and C using fuzzy_equal. If they are equal then we know that we're in one of these cases:
  * There is no HPP (most likely)
  * id parameter is not used to render the response (likely, but we already covered that before)
  * There is HPP and it's taking the first (123) value. Precedence: first. (unlikely but possible)
 * Compare A and B using fuzzy_equal. If they are equal then we know that we're in one of these cases:
  * There is no HPP (most likely)
  * id parameter is not used to render the response (likely, but we already covered that before)
  * There is HPP and it's taking the second/injected (122) value. Precedence: last. (unlikely but possible)

Note that we fill the `id` parameter with a number because it was originally filled with one. The number used for the test is the original number -1. When the parameter we're fuzzing is a string we can just use REFLECT_TOKEN

### Client-side HTTP Parameter Pollution vulnerability
From https://www.acunetix.com/blog/whitepaper-http-parameter-pollution/ , we can find this example:

> `http://host/viewemailn.jsp?client_id=79643215%26action%3Ddelete`

> After the creation of the malicious link, the page now contains two links which are injected with an extra action parameter. Thus:

> `<a href=viewemail.jsp?client_id=79643215&action=delete&action=view > View </a>`
> `<a href=viewemail.jsp?client_id=79643215&action=delete&action=delete > Delete </a>`

It seems that at least this type of HPP, where detection is done via analysing the links in the response, is not being done. One of the things I like about this method is that it only requires us to send one HTTP request per QS parameter for detection. This detection method/vulnerability check should be added as a different method in the hpp class. It's the same idea as the one shown in [page 17 of this slide deck](http://www.slideshare.net/embyte/http-parameter-pollution-vulnerabilities-in-web-applications-black-hat-eu-2011)

### HTTP precedence detection

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,259 @@
...
+from w3af.core.data.kb.info_set import InfoSet
+
+
+REFLECT_TOKEN = 'w3afHPPReflectToken'
+FORGE_TOKEN = 'w3afHPPForgeToken'
+EQUALITY_THRESHOLD = 0.95
+#paramater precedence types
+LAST = 'last occurance'
+FIRST = 'first occurance'
+CONCAT = 'concatination'
```
andresriancho: concatenation

nixwizard:
Hi, Andres. This is really a huge help, thanks!
But as for the client side hpp detection, this is what I implemented(or 
tried to) in these lines:
https://github.com/nixwizard/w3af/blob/hpp/w3af/plugins/audit/hpp.py#L176-L196
only
https://github.com/nixwizard/w3af/blob/hpp/w3af/plugins/audit/hpp.py#L194
should be:
r = self.find_reflection(vtest_dp, precedence, FORGE_TOKEN)
if we take php for example:
then one created mutants will send:
http://phptarget/test.php?a=w3afHPPReflectToken%26client_id%3Dw3afHPPForgeToken&id=1234
because w3af encodes w3afHPPReflectToken&client_id=w3afHPPForgeToken
to
w3afHPPReflectToken%26client_id%3Dw3afHPPForgeToken
before sending it on
and
response will be:
<a 
href="viewemail.php?client_id=1234&action=w3afHPPReflectToken&client_id=w3afHPPForgeToken"> 
View </a>'
So line
call to self.find_reflection 
https://github.com/nixwizard/w3af/blob/hpp/w3af/plugins/audit/hpp.py#L194
does actual link analysis
and
client side HPP vulnerability will be identified, because the precedence 
is last.

Maybe I'm missing the point, but isn't it equal to we can see in the 
acunetix paper?

andresriancho:
Ok, now that you mention it, it might be like that and I was confused by:
 * Long method
 * Mixing server and client side hpp in the same method
 * Having sections for client, then server, then client again in the same method

In other words, split things into smaller and more easier to follow chunks so the dumb reviewer can understand :sunglasses: :dancer:

andresriancho:
Closing this PR because a lot of changes need to be made to the code, and reviewing a second commit in this PR would be difficult. When sending the next PR please reference this in the comment just for me to remember.

Thanks!

PR #9599, opened 2015-04-23
Allow to use custom safe chars in querystring

kyprizel:
This patch allows to use custom safe chars in URL querystring.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -92,3 +92,4 @@ debian/w3af.substvars
 debian/w3af/
 docker/kali-debootstrap
 *.deb
+/.idea/
--- a/w3af/core/data/dc/query_string.py
+++ b/w3af/core/data/dc/query_string.py
@@ -34,7 +34,9 @@ class QueryString(KeyValueContainer):
 
     :author: Andres Riancho (andres.riancho@gmail.com)
     """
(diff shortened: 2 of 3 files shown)


andresriancho:
All right, I went through your changes and at first I thought they were a great idea but after some analysis they still need some improvements. Here are some comments:

 * Don't send changes to `.gitignore`
 * With your changes some unittests don't PASS `nosetests w3af/core/data/parsers/tests/test_url.py w3af/core/data/dc/tests/test_query_string.py`

```
$ nosetests -A 'not ci_fails' w3af/core/data/parsers/tests/test_url.py w3af/core/data/dc/tests/test_query_string.py

.................F..F..F....F........................................F.......................................

...

FAILED (failures=5)
```

The failed tests are related with the changes you've introduced. If you run the same tests in develop only three fail. I recommend you try to understand what's wrong: the tests / the code you've introduced and we should discuss here.

The branch I'm working on, which you should checkout into your repo, work there and send a new pull request (which points to safe-chars, not to andresriancho/w3af/master) is https://github.com/andresriancho/w3af/tree/safe-chars

kyprizel:
Thanks, got the problem, goint to create new pull-reques.

PR #10658, opened 2015-06-23
HTTP Parameter Pollution detection plugin with tests

nixwizard:
Only client-side vulns detection by now.
Almost all of the comments from previous PR were taken into account.

### DIFF
--- a/w3af/plugins/audit/hpp.py
+++ b/w3af/plugins/audit/hpp.py
@@ -1,0 +1,263 @@
+"""
+hpp.py
+ 
+Copyright 2015 Andres Riancho
+ 
+This file is part of w3af, http://w3af.org/ .
+ 
+w3af is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation version 2 of the License.
+ 
+w3af is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+ 
+You should have received a copy of the GNU General Public License
+along with w3af; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+"""
+
+from w3af.core.controllers.plugins.audit_plugin import AuditPlugin
+from w3af.core.data.fuzzer.fuzzer import create_mutants
+from w3af.core.data.kb.vuln import Vuln
+from w3af.core.data.kb.info import Info
+from w3af.core.data.fuzzer.mutants.querystring_mutant import QSMutant
+import w3af.core.controllers.output_manager as om
+import w3af.core.data.constants.severity as severity
+from w3af.core.controllers.misc.fuzzy_string_cmp import fuzzy_equal
+import w3af.core.data.parsers.parser_cache as parser_cache
+from lxml import etree
+from w3af.core.controllers.exceptions import BaseFrameworkException
+from w3af.core.data.parsers.url import URL
+
+REFLECT_TOKEN = 'w3afHPPReflectToken'
+FORGE_TOKEN = 'w3afHPPForgeToken'
+EQUALITY_THRESHOLD = 0.95
+# paramater precedence types
+LAST = 'last occurance'
+FIRST = 'first occurance'
+CONCAT = 'concatenation'
+
+
+def get_forgeable(reflection, prec):
+    #this function returns list of querystring parameters which may be
+    #forged provided precedence is prec
+    r = []
+    param, url = reflection
+    qs = url.get_querystring()
+
+    passed = 0
+    for p in qs:
+        if p == param:
+            passed = 1
+            continue
+        if passed == 0 and prec == LAST:
+            r += [p]
+        if passed == 1 and prec == FIRST:
+            r += [p]
+    return r
+
+
+def find_reflection(response, precedence, patterns):
+    #finds reflection of a string or two concatenated
+    # strings in web page links or input parameters
+    if len(patterns) > 1:
+        pattern1, pattern2 = patterns
+    else:
+        pattern1, pattern2 = patterns[0], None
+
+    try:
+        parser = parser_cache.dpc.get_document_parser_for(response)
+    except BaseFrameworkException:
+        return
+
+    pattern1 = unicode(pattern1)
+    result = []
+    #find reflection in links
+    for references in parser.get_references():
+        for ref in references:
+            if not ref:
+                continue
+            qs = ref.get_querystring()
+            for p in qs:
+                if precedence == LAST and len(qs[p]) == 2:
+                    rv = qs[p][1]
+                else:
+                    rv = qs[p][0]
+                #looking for single reflection if only one pattern was
+                # entered
+                if rv == pattern1 and not pattern2:
+                    result += [(p, ref)]
+
+                if not pattern2:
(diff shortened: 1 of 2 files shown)


w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+
+
+class hpp(AuditPlugin):
+    """
+    Identify client-side http parameter pollution vulnerabilities.
+    """
+
+    def __init__(self):
+        AuditPlugin.__init__(self)
+        self.parser = etree.HTMLParser(remove_blank_text=True, remove_comments=True)
```
andresriancho: Not used, remove

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+        :author: Dmitry Roshchin (nixwizard@gmail.com)
+        """
+
+        if not orig_response.is_text_or_html():
+            return
+
+        if not orig_response.get_code() == 200:
+            return
+
+        method = freq.get_method()
```
andresriancho: No need to define this variable since it's only used once

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+
+        if method == 'GET':
+            qs = freq.get_querystring()
+
+        precedence = None
+        #finding reflected params if any
+        for param in qs:
+            value = qs[param][0]
+            #checking if the value is number or string so that our
+            #reflection test request would be more legitimate
```
andresriancho: I like this approach, I'm in doubt how you'll match later, we'll see ;)

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+
+        method = freq.get_method()
+
+        if method == 'GET':
+            qs = freq.get_querystring()
+
+        precedence = None
+        #finding reflected params if any
+        for param in qs:
+            value = qs[param][0]
```
andresriancho: There might be more than one query string parameter with the same name, for example:

`http://foo.com/?q=abc&q=def`

By using `qs[param][0]` you're only testing for HPP in the first `q`.

On a related note, what would happen if the app has two `q` params "by default" (without any fuzzing), will the plugin go crazy?

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+                                      value + '&' + param + '=' + newval))
+            ptest_resp = self._uri_opener.GET(ptest_url)
+            ptest_resp_code = ptest_resp.get_code()
+
+            if not ptest_resp_code == 200:
+                om.out.debug(u"Precedence check request to the uri %s "
+                             u"failed! "
+                             u"Got response code %s in response id %s\n"
+                             u"Unable to continue HPP detection. "
+                             u"Resource might be protected by a WAF,"
```
andresriancho: `WAF,` should be `WAF.`

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+from w3af.core.controllers.plugins.audit_plugin import AuditPlugin
+from w3af.core.data.fuzzer.fuzzer import create_mutants
+from w3af.core.data.kb.vuln import Vuln
+from w3af.core.data.kb.info import Info
+from w3af.core.data.fuzzer.mutants.querystring_mutant import QSMutant
+import w3af.core.controllers.output_manager as om
+import w3af.core.data.constants.severity as severity
+from w3af.core.controllers.misc.fuzzy_string_cmp import fuzzy_equal
+import w3af.core.data.parsers.parser_cache as parser_cache
+from lxml import etree
```
andresriancho: Remove, not used

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+        """
+        :return: A DETAILED description of the plugin functions and features.
+        """
+        return """
+        This plugin detects client-side HTTP Parameter Pollution
+        vulnerabilities using the following technique. First it detects if any
+        of the input params is being reflected. If value of such parameter
+        exists in any of the links of received HTML code it will try to forge
+        it using HPP attack according to determined precedence or just
+        report if precedence is concatenation.
```
andresriancho: This plugin detects client-side HTTP Parameter Pollution vulnerabilities

The following technique is used:
 * Detect if any of the input params is being reflected
 * For each reflected parameter check if they are used in HTML links
 * Send an HPP check to inject an extra parameter in the link

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+
+                vtest_resp = self._uri_opener.send_mutant(mu)
+                #checkig for successful exploitation
+                r = find_reflection(vtest_resp, precedence, [forged_val])
+                if r:
+                    msg = u"The %s uri is vulnerable to HPP attack: value of " \
+                          u"parameter %s can be forged using URL: %s, " \
+                          u"precedence is %s"
+                    msg = msg % (freq.get_uri(), r, mu.get_uri(), precedence)
+                    v = Vuln.from_mutant(u'HPP', desc=msg, mutant=mu,
```
andresriancho: Use full name for vuln, the user might not know what "HPP" is and can't google that. Use "HTTP Parameter Pollution" instead.

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+            if orig_response.body == rtest_resp.body:
+                #parameter param is not being reflected going for the next
+                continue
+
+            #Now we are doing precedence test
+            u = str(mu.get_uri())
+            #this hack is to make w3af send request with unencoded qs
+            #because as long as we are trying to use qs it urlencodes the
+            #second instance of param with new value
+            ptest_url = URL(u.replace(newval,
```
andresriancho: What if you're testing `http://foo122.com/?id=121` ?

That would replace the domain too!

w3af/plugins/audit/hpp.py
```diff
@@ -0,0 +1,263 @@
...
+                             u"Got response code %s in response id %s\n"
+                             u"Unable to continue HPP detection. "
+                             u"Resource might be protected by a WAF,"
+                             % (ptest_url, ptest_resp_code, ptest_resp.id))
+                #no need to continue with the param if it is not 200 OK
+                continue
+
+            ptest_body = ptest_resp.body
+
+            if fuzzy_equal(orig_response.body,
```
andresriancho: Do we really need the precedence check for client-side? Isn't this too complex? Why not just set each input parameter to `value&param=3` , then parse the response for links, and check if the links have the extra-unexpected `param` parameter?

I'm getting really confused with all this...

nixwizard: I believe precedence check is essential for confirming HPP. Here is the reason.
We could input do value&param=3 as you say. But we cannot confirm
vulnerability without knowing precedence, for example we have freq

http://test.com/?name=john

and result link in html code: 

http://test.com/manageuser/?name=john&action=read

so we are checking reflection and see that id is reflected.
If then we just do http://test.com/manageuser/?name=john%26action%3Ddelete
(like ZAP does for example)
we get link:
http://test.com/manageuser/?name=john&action=delete&action=read
we can conclude that we've detected potential vulnerability.
But can we exploit it? Can we pollute paramtter "action"? 
It depends on precedence:
If it is FIRST - yes we can.
if it is LAST - no, result action value will always be "read" in this case.
So if we do not know precedence the plugin generates lots of
false positives. And we supply our users with more valuable
data. Agree?

Issue #10615, opened 2015-06-23
Add basic authentication to REST API


andresriancho:
## Looking for contributor!
Want to improve the w3af framework? This is your chance! This task is looking for a contributor, just add a message to this issue saying: "I'll do it!" and start with the fun :+1: 

## User story
As a user I want to restrict access to the REST API in order to be able to access it over the network without any risk of others accessing it

## Conditions of satisfaction
 - [x] Decorate all routes with the basic authentication checker
 - [x] Let the user define username and password using command line arguments and a yaml configuration file (to prevent "password sniffing using `ps auxf`). The yaml file can also be used to configure bind port and address
 - [x] Store SHA-512 password hashes in yaml
 - [x] Require the user to enter SHA-512 hash of the password in command line
 - [x] Answer with a 401 for all requests which don't have the right credentials
 - [x] Write documentation on how to provide authentication, consume authenticated service, etc.
 - [ ] Allow the user to start API without authentication

## Limitations
 - [x] Only one user is allowed

OwenTuz:
I'll do this one :)

andresriancho:
@OwenTuz amazing, thanks!

andresriancho:
I just added some comments to the code, please also remember to add the user documentation to https://github.com/andresriancho/w3af/blob/master/doc/sphinx/api/index.rst explaining how to start the API with auth

andresriancho:
Your code looks really good for now!

OwenTuz:
I've attached an initial pull request #10857 above (please ignore 10856, was for the wrong branch - sorry), with some notes. Let me know if you want me to change anything, or spot any bugs/problems.

Thanks!

andresriancho:
@OwenTuz amazing code! A couple of fixes and it will be ready to merge

PR #10856, opened 2015-06-30
Feature/api basic auth

OwenTuz:
#10615 - Add basic auth to REST API
- Adds @requires_auth decorator to API, which checks basic auth against a SHA512-hashed password
- Adds CLI options for 'username' and 'password'
- Adds YAML configuration option for API (using CLI option "-c /path/to/file.yaml")
- Updates API tests, Docker API script and documentation to reflect the change

Note that this breaks backwards compatibility by forcing a requirement for authentication on all views (ie users must specify a password hash either on the CLI or in a configuration file - there's no default). It also requires users running the Docker script to specify a configuration file containing at minimum their hashed password.

I'm happy to refactor that if we would prefer to make basic authentication optional and retain backward compatibility.

### DIFF
--- a/doc/sphinx/api/index.rst
+++ b/doc/sphinx/api/index.rst
@@ -55,6 +55,14 @@ For example, using the 'curl' command:
       "docs": "http://docs.w3af.org/en/latest/api/index.html" 
     }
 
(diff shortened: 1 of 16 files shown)


OwenTuz:
...wrong branch

PR #12036, opened 2015-08-10
Add wad plugin

plizonczyk:
This should close https://github.com/andresriancho/w3af/issues/1081

This PR adds new infrastructure plugin, which runs [WAD](https://github.com/CERN-CERT/WAD) tool. It uses [Wappalyzer's](https://github.com/AliasIO/Wappalyzer) rules database to detect technologies used websites, ranging from webservers, through frameworks, to javascript plugins.

As this is my first PR to this project, I'm eager for your extensive code review.


w3af/core/controllers/dependency_check/requirements.py
```diff
@@ -52,6 +52,7 @@
                      PIPDependency('jinja2', 'Jinja2', '2.7.3'),
                      PIPDependency('vulndb', 'vulndb', '0.0.17'),
                      PIPDependency('markdown', 'markdown', '2.6.1'),
+                     PIPDependency('wad', 'wad', '0.2.0'),
```
andresriancho: https://github.com/andresriancho/w3af-kali/issues/10

andresriancho:
This PR will be rejected since it was made against `master`. Please send it again but against the `develop` branch.

I'll review the code anyways so the next PR is cleaner and merged fast :+1:

w3af/plugins/infrastructure/webapp_detection.py
```diff
@@ -0,0 +1,139 @@
...
+
+from w3af.core.controllers.plugins.infrastructure_plugin import InfrastructurePlugin
+from w3af.core.controllers.exceptions import RunOnce
+from w3af.core.controllers.misc.decorators import runonce
+from w3af.core.data.kb.info import Info
+from w3af.core.data.options.opt_factory import opt_factory
+from w3af.core.data.options.option_list import OptionList
+
+
+class w3afDetector(Detector):
```
andresriancho: Awesome, this is what I meant by subclassing the detector class.

w3af/plugins/infrastructure/webapp_detection.py
```diff
@@ -0,0 +1,139 @@
...
+    def get_new_url(self, page):
+        return page.get_url().url_string
+
+    def get_content(self, page, url):
+        return page.get_body()
+
+
+class webapp_detection(InfrastructurePlugin):
+    """
+    Identify technologies used by website based on HTTP response and HTML content
```
andresriancho: Please check what happens when you do this:

```
./w3af_console
plugins
infrastructure list
```

This class doc is used in that display. Usually all plugins have one line in the doc which is displayed, not sure if something will break with more lines.

w3af/plugins/infrastructure/webapp_detection.py
```diff
@@ -0,0 +1,139 @@
...
+    def discover(self, fuzzable_request):
+        """
+        It calls the "main" from WAD and writes the results to the kb.
+
+        :param fuzzable_request: A fuzzable_request instance that contains
+                                    (among other things) the URL to test.
+        """
+        msg = 'Web application detection (WAD) plugin is starting. ' \
+              'Original WAD authors: CERN-CERT; http://cern.ch/security'
+        om.out.information(msg)
```
andresriancho: If all plugins would send a log like this to information() the log would have a lot of garbage. Please remove these two lines.

You can add the original wad authors to the plugin description

w3af/plugins/infrastructure/webapp_detection.py
```diff
@@ -0,0 +1,139 @@
...
+
+        self._main(fuzzable_request.get_url())
+
+    def _main(self, url):
+        """
+        Based on WAD's main executable
+        """
+
+        detector = w3afDetector(uri_opener=self._uri_opener)
+        results = detector.detect(url, timeout=self._timeout)
```
andresriancho: w3af's urlopener already has a default timeout which is set via `http-settings`, I believe there is no value added to the user if we provide a setting to override it in this plugin. Please remove the `self._timeout` and the option to configure it.

w3af/plugins/infrastructure/webapp_detection.py
```diff
@@ -0,0 +1,139 @@
...
+    def _report(self, results):
+        """
+        Displays detailed report information to the user and save the data to
+        the kb.
+
+        :return: None.
+        """
+        if results:
+            i = Info('Web applications detection report',
+                     HumanReadableOutput().retrieve(results), 1,
```
andresriancho: Please replace the 1 with the real HTTP request ID used to identify this information

w3af/plugins/tests/infrastructure/test_webapp_detection.py
```diff
@@ -0,0 +1,59 @@
...
+    _run_configs = {
+        'cfg': {
+            'target': target_url,
+            'plugins': {'infrastructure': (PluginConfig('webapp_detection'),)}
+        }
+    }
+
+    MOCK_RESPONSES = [
+        MockResponse('http://webapp-detection/jquery-in-content', '<script src="jquery.js"></script>'),
+        MockResponse('http://webapp-detection/apache-in-headers-no-content', '', headers={'Server': 'Apache'}),
```
andresriancho: Please respect 80 column limit

w3af/plugins/tests/infrastructure/test_webapp_detection.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_apache(self):
+        cfg = self._run_configs['cfg']
+        self._scan(self.target_url + 'apache-in-headers-no-content', cfg['plugins'])
+
+        infos = self.kb.get('webapp_detection', 'webapp_detection')
+
+        self.assertEqual(len(infos), 1, infos)
+        info = infos[0]
+
+        self.assertIn('Apache', info.get_desc())
```
andresriancho: Also assert that jquery is not in desc

w3af/plugins/tests/infrastructure/test_webapp_detection.py
```diff
@@ -0,0 +1,59 @@
...
+    def test_jquery(self):
+        cfg = self._run_configs['cfg']
+        self._scan(self.target_url + 'jquery-in-content', cfg['plugins'])
+
+        infos = self.kb.get('webapp_detection', 'webapp_detection')
+
+        self.assertEqual(len(infos), 1, infos)
+        info = infos[0]
+
+        self.assertIn('jQuery', info.get_desc())
```
andresriancho: Also assert that apache is not in desc

w3af/plugins/infrastructure/webapp_detection.py
andresriancho: This saves the findings to the KB in such a way that it's readable for humans, which is great and will cover 95% of the cases, but something else that we usually do is to `raw_write` data like this in a way that can be easily read/consumed by other plugins.

Not sure how the wad data looks like, but if you could save a dict (or a list, or something similar) containing:
```python
{'jquery': '1.2.3',
'apache': 'unknown'}
```

That would be amazing :)

See: `find . -name '*.py' | xargs grep 'raw_write' | grep -v 'test_'`

andresriancho:
Overall, code and tests look really good, great job!

w3af/plugins/infrastructure/webapp_detection.py
```diff
@@ -0,0 +1,139 @@
+"""
+webapp_detection.py
```
andresriancho: The plugin name could be improved, maybe "application_fingerprint" ?

w3af/plugins/infrastructure/webapp_detection.py
```diff
@@ -0,0 +1,139 @@
...
+
+    def _report(self, results):
+        """
+        Displays detailed report information to the user and save the data to
+        the kb.
+
+        :return: None.
+        """
+        if results:
+            i = Info('Web applications detection report',
```
andresriancho: Change the title to `Application fingerprint`

w3af/plugins/infrastructure/webapp_detection.py
```diff
@@ -0,0 +1,139 @@
...
+
+    :author: Piotr Lizonczyk (piotr.lizonczyk@gmail.com)
+    """
+    def __init__(self):
+        InfrastructurePlugin.__init__(self)
+
+        # User configured
+        self._timeout = 5
+
+    @runonce(exc_class=RunOnce)
```
andresriancho: The `RunOnce` is something I would also use, but I was wondering... is this the best we can do? Let me give you a couple of examples:

 * The first fuzzable request that arrives to this plugin is usually http://foo.com/ , in some cases that might just be a redirect to something else like http://foo.com/home/ . w3af doesn't automatically follow redirects, so wad would analyze an empty result

 * http://foo.com/ is just a simple page with links to the different application sections, / doesn't use any JS frameworks nor CMS/etc, but the deeply linked apps do.

In both those cases WAD could find stuff and we're not invoking it to do so.

What if we let w3af call WAD several times until no more new information is collected and then we raise `NoMoreCalls`? Broken pseudo-code:

```python
url_knowledge = self.analyze_url(url)
if self.is_new_knowledge(url_knowledge):
    self.wad_knowledge.merge(url_knowledge)
    self.no_new_knowledge_count = 0
else:
    self.no_new_knowledge_count += 1
    if self.no_new_knowledge_count == 10:
        raise NoMoreCalls
```

I randomly chose `10`, you can change that to any value based on your previous experience with wad

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: Nothing bad actually happened, lines were concatenated, yet I've changed it.

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: Sure, I've patterned it on Halberd plugin.

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: :+1:

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: ID fixed, I've spotted it hardcoded in mulitple other plugins (I'll later post an issue about that).

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: Fine.

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: Raw dict looks like this:
```python
{
    "http://w3af.org": [
        {"type": "web-servers",  "app": "Nginx", "ver": None}, 
        {"type": "cdn", "app": "CloudFlare", "ver": None},
        {"type": "cms,blogs", "app": "WordPress", "ver": None}, 
        {"type": "javascript-frameworks", "app": "jQuery", "ver": None},
        {"type": "photo-galleries,javascript-frameworks", "app": "prettyPhoto", "ver": None}, 
        {"type": "font-scripts", "app": "Font Awesome", "ver": None}, 
        {"type": "font-scripts", "app": "Google Font API", "ver": None}, 
        {"type": "programming-languages", "app": "PHP", "ver": None}  
    ]
}
```

If it's fine, I can easily include it as a switchable option (human/raw data). Do you suggest changing the format to the one that you provided?

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: Yeah, that's what should be done, I didn't know the mechanism. Should every result be logged in as another Info in knowledge base? If not, then which HTTP response ID should I log?

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: Alright, I see that it doesn't need to be option, I'll simply store both. :)

w3af/plugins/infrastructure/webapp_detection.py
andresriancho: :+1:

w3af/plugins/infrastructure/webapp_detection.py
andresriancho: > Should every result be logged in as another Info in knowledge base?

No, I would log it as one Info with all the merged information. Just write to the `kb` before `raise NoMoreCalls`

> If not, then which HTTP response ID should I log?

You can send a list of IDs in the parameter with all the http responses you've used/merged to create that Info.

w3af/plugins/infrastructure/webapp_detection.py
andresriancho: > I've spotted it hardcoded in mulitple other plugins (I'll later post an issue about that).

Thanks!

andresriancho: > If it's fine, I can easily include it as a switchable option (human/raw data). 

No, you should write to the KB in both formats every time. For reports, GUI, etc. only the Info sub-classes are shown to the user so don't worry about how/if this dict will show to the user.

> Do you suggest changing the format to the one that you provided?

The existing format is OK, keep it please.

plizonczyk:
Closing this MR (bad target branch https://github.com/andresriancho/w3af/pull/12036#issuecomment-129501778)

w3af/plugins/infrastructure/webapp_detection.py
plizonczyk: >>Should every result be logged in as another Info in knowledge base?

>No, I would log it as one Info with all the merged information. Just write to the kb before raise NoMoreCalls

It isn't really possible - we may never reach maximum no_new_knowledge_count. The solution I can come up with is to update the information in KB on every run of the plugin.

PR #12208, opened 2015-08-14
Add wad plugin

plizonczyk:
(This branch was checked out from master, but some commits from master weren't merged into develop, so they show up in MR)
(Some functionality is yet to be added (configurable amount of max retries without new knowledge))

@andresriancho I think I've addressed all your comments from original merge request. Although I've ran into problems with new version of code (not using RunOnce).
I assume, that plugin can be ran simultaneously for e.g. two urls at the same time. This, unfortunately creates some race conditions (I guess that is the issue). I've ran the same profile (with application_fingerprint plugin only) on same 2 URLs. The example outcomes are:
1) Full success, no error messages, proper IDs of requests 
2) No error messages, yet KB entry contains results from only one URL
3) `A "AttributeError" exception was found while running infrastructure.application_fingerprint on "Method: GET | http://domain/path/foo/". The exception was: "'NoneType' object has no attribute 'id'" at application_fingerprint.py:discover():89. The scan will continue but some vulnerabilities might not be identified.`, but the KB entry contains proper results, though with only one request ID
4) No error messages, but duplicated IDs of requests in report

The question is, is there a w3af-specific way of ensuring thread-safeness of code (like proper decorators that will lock the thread), or should I just try to rewrite the code?


w3af/plugins/infrastructure/application_fingerprint.py
```diff
@@ -0,0 +1,149 @@
...
+
+        self._report()
+        if self.no_new_knowledge_count >= self.no_new_knowledge_count_max:
+            raise NoMoreCalls
+
+    def _main(self, fuzzable_request):
+        """
+        Based on WAD's main executable
+        """
+        self.detector = w3afDetector(uri_opener=self._uri_opener)
```
andresriancho: I believe that this is what's causing the thread-unsafe code, `self.detector` is being shared between multiple threads. What if you replace `self.detector` with `detector`? (create one instance of that object for each thread run, it doesn't sound very expensive to do it like that)

w3af/plugins/infrastructure/application_fingerprint.py
```diff
@@ -0,0 +1,149 @@
...
+        :param fuzzable_request: A fuzzable_request instance that contains
+                                    (among other things) the URL to test.
+        """
+        result = self._main(fuzzable_request)
+        if not result:
+            return
+
+        self.knowledge.update(result)
+        self.full_results.update(result)
+        self.request_ids.append(self.detector.original_response.id)
```
andresriancho: > duplicated IDs of requests in report

You might fix this by using a `set()` instead of a list for `self.request_ids`

w3af/plugins/infrastructure/application_fingerprint.py
andresriancho: The previous change would also require a minor change in `discover`, but nothing you can't handle :+1:

w3af/plugins/infrastructure/application_fingerprint.py
```diff
@@ -0,0 +1,149 @@
...
+        the kb.
+
+        :return: None.
+        """
+        desc = HumanReadableOutput().retrieve(self.full_results)
+        i = Info('Application fingerprint',
+                 desc, self.request_ids, self.get_name())
+
+        if not self.kb_info_id:
+            kb.kb.append(self, 'Application fingerprint', i)
```
andresriancho: Please use `application_fingerprint` instead of `Application fingerprint` here

w3af/plugins/infrastructure/application_fingerprint.py
plizonczyk: Actually this might have been caused by shared detector instance.

andresriancho:
Code looks really great, with the thread-safe fixes I propose it will be ready to merge.

PS: Note that even if the code is 100% ready I won't merge it into develop right away since I'm still working on #10821 and I don't want to merge a change that requires a new kali package to be added (that would just slow down the kali release).

plizonczyk:
Fixed all non-thread safe statements in code. Also, I had to wrap instatiation of Detector class with lock, since it's not thread-safe in WAD right now (first instatiation prepares rules DB for use and it breaks if ran simultaneously). I'll work on making it threadsafe in WAD and update appropriately.

Thanks for your assistance @andresriancho

andresriancho:
As a final step I always run the PR locally to check everything is running fine, during this step I found that the plugin is not showing anything to the user via the output manager; this means that a user that's running the plugin using the console-ui won't see anything in the output. Try it yourself:

Write this content to `test.w3af`:
```text
plugins infrastructure application_fingerprint
plugins output console
plugins output config console
set verbose True
back


target
set target http://www.clarin.com/
back

start

exit
```

And then `./w3af_console -s test.w3af`

This is fixed by adding:

```python
    def end(self):
        """
        Required in order to print the application fingerprint to the console
        only once (at the end of the scan).
        
        :return: None
        """
        if self.kb_info_id:
            info = kb.kb.get_by_uniq_id(self.kb_info_id)
            om.out.report_finding(info)
```

You'll have to import `import w3af.core.controllers.output_manager as om`

Once that was fixed I was able to identify two more cosmetic bugs in the output:
```text
Web application detection results
Results for website http://www.tagcube.io/, found applications:
	Nginx (web-servers), version: 1.6.2
	jQuery (javascript-frameworks), version: 1.10.2
	Underscore.js (javascript-frameworks)
	AngularJS (javascript-frameworks)
	Twitter Bootstrap (web-frameworks)
	Font Awesome (font-scripts)
	Google Font API (font-scripts)

. This information was found in the request with id 34.
```

The first one is in `HumanReadableOutput`:

```
Web application detection results
Results for website http://www.tagcube.io/, found applications:
```

I believe that those two lines should be changed to:
```
Web application detection results for website http://www.tagcube.io/ :
```

Then the end is ugly:
```
. This information was found in the request with id 34.
```

I'll fix this myself in https://github.com/andresriancho/w3af/issues/12220

andresriancho:
PS: Sorry for all the small details

plizonczyk:
> PS: Sorry for all the small details
If we chase perfection, we can catch excellence - namely, I don't mind :+1: 

I'll fix up output along with threading in WAD.

plizonczyk:
I've updated version of WAD in dependencies, it has some bugfixes and is thread-safe.

andresriancho:
Awesome, thanks!

Still working on the w3af.deb , mainly blocked by kali 2.0 not being released at https://hub.docker.com/u/kalilinux/

andresriancho:
Closing this PR, see comment in https://github.com/andresriancho/w3af/issues/1081

PR #12449, opened 2015-08-30
SSL Support For the API

breakthesec:
I've added the SSL Support for the API.  You can place the ssl key & certificate in the root directory of w3af (Eg: w3af/ssl/w3af.cert)

### DIFF
--- a/w3af/core/ui/api/main.py
+++ b/w3af/core/ui/api/main.py
@@ -24,7 +24,7 @@ import argparse
 
 from w3af.core.ui.api import app
 from w3af.core.ui.api.utils.cli import process_cmd_args_config
-
+from w3af.core.ui.api.utils.digital_certificate import SSLCertifiacate
 
 def main():
     """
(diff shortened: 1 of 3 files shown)


w3af/core/ui/api/main.py
```diff
@@ -39,8 +40,16 @@ def main():
...
     # And finally start the app:
     try:
-        app.run(host=app.config['HOST'], port=app.config['PORT'],
-                debug=args.verbose, use_reloader=False, threaded=True)
+        if args.disableSSL:
+            app.run(host=app.config['HOST'], port=app.config['PORT'],
+                    debug=args.verbose, use_reloader=False, threaded=True)
+        else:
+            context = ssl.SSLContext(ssl.PROTOCOL_TLSv1_2)
+            context.load_cert_chain('ssl/w3af.cert', 'ssl/w3af.key')
```
andresriancho: So, by default the REST API doesn't start until the user either:

 * Adds `--no-ssl` to the command line
 * Creates the cert and key files and puts them in the right directory

I disagree with this decision. By default the REST API should start without any command line arguments or previous commands being run.

The best would be to have this:
 * REST API by default creates a cert and key file set, self-signed, stores them in the `~/.w3af/ssl` directory
 * The user can use `-no-ssl` to completely disable SSL
 * The user can override the defaults and define specific paths for the cert and key files

Also, we need to document all this somewhere at https://github.com/andresriancho/w3af/tree/master/doc/sphinx/api

andresriancho:
I love the feature, but the design decisions are not well thought.

breakthesec:
Yes that can be done.

andresriancho:
Awesome, please do send those improvements, I would like to have SSL support.

Final note, please send your pull request against `andresriancho:develop` instead of `andresriancho:master`

PR #12853, opened 2015-10-14
Feature/rest api allow config updates

OwenTuz:
I believe this is a working branch with tests, but there are a few things worth discussing - I'll add notes to the commits.

### DIFF
--- a/w3af/core/controllers/core_helpers/status.py
+++ b/w3af/core/controllers/core_helpers/status.py
@@ -40,18 +40,18 @@ class w3af_core_status(object):
     def __init__(self, w3af_core, scans_completed=0):
         # Store the core to be able to access the queues to get status
         self._w3af_core = w3af_core
-        
+
         # Init some internal values
         self._is_running = False
         self._paused = False
         self._start_time_epoch = None
         self.scans_completed = scans_completed
-        
+
         # This indicates the plugin that is running right now for each
         # plugin_type
         self._running_plugin = {}
         self._latest_ptype, self._latest_pname = None, None
-        
+
         # The current fuzzable request that the core is analyzing at each phase
         # where a phase means crawl/audit
         self._current_fuzzable_request = {}
@@ -75,17 +75,17 @@ class w3af_core_status(object):
         """
         if self._paused:
             return PAUSED
-        
+
         elif not self.is_running():
             return STOPPED
-        
+
         else:
             crawl_plugin = self.get_running_plugin('crawl')
             audit_plugin = self.get_running_plugin('audit')
-            
+
             crawl_fr = self.get_current_fuzzable_request('crawl')
             audit_fr = self.get_current_fuzzable_request('audit')
-            
+
             if crawl_plugin == audit_plugin == crawl_fr == audit_fr == None:
                 return 'Starting scan.'
 
@@ -97,10 +97,10 @@ class w3af_core_status(object):
             if audit_plugin is not None and audit_fr is not None:
                 if status_str:
                     status_str += '\n'
-                
+
                 status_str += 'Auditing %s using %s.%s' % (audit_fr, 'audit',
                                                            audit_plugin)
-                
+
             status_str = status_str.replace('\x00', '')
             return status_str
 
@@ -114,7 +114,7 @@ class w3af_core_status(object):
         """
         if log:
             om.out.debug('Starting plugin: %s.%s' % (plugin_type, plugin_name))
-            
+
         self._running_plugin[plugin_type] = plugin_name
         self._latest_ptype, self._latest_pname = plugin_type, plugin_name
 
@@ -137,17 +137,17 @@ class w3af_core_status(object):
         core is still working, it should call is_running() to know that.
         """
         return self._is_running
-    
+
     def is_paused(self):
         return self._paused
-    
+
     def get_run_time(self):
         """
         :return: The time (in minutes) between now and the call to start().
         """
         if self._start_time_epoch is None:
             raise RuntimeError('Can NOT call get_run_time before start().')
-        
+
         now = time.time()
         diff = now - self._start_time_epoch
         run_time = diff / 60
@@ -166,13 +166,13 @@ class w3af_core_status(object):
                  start of the scan.
         """
         if self._start_time_epoch is None:
-            raise RuntimeError('Can NOT call get_run_time before start().')
-        
+            return None
+
         now = time.time()
         diff = now - self._start_time_epoch
         run_time = diff / 60.0
         return int(consecutive_number_generator.get() / run_time)
-    
+
     def scan_finished(self):
         self._is_running = False
         self._running_plugin = {}
@@ -196,7 +196,7 @@ class w3af_core_status(object):
     def get_crawl_input_speed(self):
         dc = self._w3af_core.strategy._discovery_consumer
         return None if dc is None else round_or_None(dc.in_queue.get_input_rpm())
-    
+
     def get_crawl_output_speed(self):
         dc = self._w3af_core.strategy._discovery_consumer
         return None if dc is None else round_or_None(dc.in_queue.get_output_rpm())
@@ -227,7 +227,7 @@ class w3af_core_status(object):
 
         if input_speed is None or output_speed is None:
             return None
-            
+
         if input_speed >= output_speed:
             return None
 
@@ -244,7 +244,7 @@ class w3af_core_status(object):
     def get_audit_input_speed(self):
         ac = self._w3af_core.strategy._audit_consumer
         return None if ac is None else round_or_None(ac.in_queue.get_input_rpm())
-    
+
     def get_audit_output_speed(self):
         ac = self._w3af_core.strategy._audit_consumer
         return None if ac is None else round_or_None(ac.in_queue.get_output_rpm())
@@ -269,14 +269,14 @@ class w3af_core_status(object):
 
         if input_speed is None or output_speed is None:
             return None
-        
+
         if input_speed >= output_speed:
             return None
-        
+
         # The speed is in URLs per minute that are processed by the framework
         speed = output_speed - input_speed
         current_size = self.get_audit_qsize()
-        
+
         eta_minutes = current_size / speed
         # TODO: Show this in h/m/s
         return '%s minutes' % eta_minutes
@@ -346,35 +346,35 @@ class w3af_core_status(object):
     def get_long_status(self):
         if not self.is_running():
             return self.get_status()
-        
+
         data = {
                 'status': self.get_status(),
-                
+
                 'cin': self.get_crawl_input_speed(),
                 'cout': self.get_crawl_output_speed(),
                 'clen': self.get_crawl_qsize(),
                 'ceta': self.get_crawl_eta(),
-                
+
                 'ain': self.get_audit_input_speed(),
                 'aout': self.get_audit_output_speed(),
                 'alen': self.get_audit_qsize(),
                 'aeta': self.get_audit_eta(),
-                
+
                 'rpm': self.get_rpm()
                 }
-        
+
         status_str = '%(status)s\n'
-        
+
         status_str += 'Crawl phase: In (%(cin)s URLs/min)'\
                       ' Out (%(cout)s URLs/min) Pending (%(clen)s URLs)'\
                       ' ETA (%(ceta)s)\n'
-                      
+
         status_str += 'Audit phase: In (%(ain)s URLs/min)'\
                       ' Out (%(aout)s URLs/min) Pending (%(alen)s URLs)'\
                       ' ETA (%(aeta)s)\n'
-                      
+
         status_str += 'Requests per minute: %(rpm)s'
-        
+
         return status_str % data
 
 
--- a/w3af/core/ui/api/resources/scans.py
+++ b/w3af/core/ui/api/resources/scans.py
@@ -64,7 +64,12 @@ def start_scan():
     # REST API is an MVP and we can only run one scan at the time (for now)
     #
     scan_infos = SCANS.values()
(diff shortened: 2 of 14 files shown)


andresriancho:
Awesome PR! I believe it's one of the most complex PRs w3af has ever received. Starting my review.

w3af/core/controllers/core_helpers/status.py
```diff
@@ -166,13 +166,13 @@ def get_rpm(self):
                  start of the scan.
         """
         if self._start_time_epoch is None:
-            raise RuntimeError('Can NOT call get_run_time before start().')
-        
+            return None
```
andresriancho: Hmmm, why this change? I believe that the exception was correct, but if you've got a good reason we can keep it.

w3af/core/ui/api/resources/scans.py
```diff
@@ -64,9 +64,14 @@ def start_scan():
     # REST API is an MVP and we can only run one scan at the time (for now)
     #
     scan_infos = SCANS.values()
-    if not all([si is None for si in scan_infos]):
+    if not all([
+        (
+        si is None or
+        si.w3af_core.status.get_simplified.status.upper == 'STOPPED'
```
andresriancho: Should be `get_simplified_status().upper()` ?

w3af/core/ui/api/resources/sessions.py
```diff
@@ -0,0 +1,343 @@
...
+                                             enable_or_disable_plugin)
+from w3af.core.ui.api.db.master import SCANS
+from w3af.core.ui.api.utils.scans import (get_scan_info_from_id,
+                                          create_scan_helper,
+                                          start_preconfigured_scan)
+from w3af.core.controllers.w3afCore import w3afCore
+from w3af.core.controllers.exceptions import BaseFrameworkException
+from w3af.core.controllers.misc_settings import MiscSettings
+from w3af.core.data.options.opt_factory import opt_factory
+from w3af.core.data.options.option_types import *
```
andresriancho: Please avoid wildcard imports

w3af/core/ui/api/resources/sessions.py
```diff
@@ -0,0 +1,343 @@
...
+        - The newly created scan ID (eg. 0)
+    """
+    scan_id = create_scan_helper()
+
+    return jsonify({'message': 'Success',
+                    'id': scan_id,
+                    'href': '/sessions/%s' % scan_id}), 201
+
+
+@app.route('/sessions/<int:scan_id>', methods=['GET'])
```
andresriancho: Just a naming thing but... `int:scan_id` should be `int:session_id`?

w3af/core/ui/api/resources/sessions.py
```diff
@@ -0,0 +1,343 @@
...
+
+    :return: 404 if scan does not exist.
+             Otherwise, a list of possible routes under /sessions/<scan_id>/
+    """
+    urls = list_subroutes(request)
+    return jsonify({'message': 'Session %s' % scan_id,
+                    'available_endpoints': urls})
+
+
+@app.route('/sessions/<int:scan_id>/plugins/', methods=['GET'])
```
andresriancho: Same as above with `int:session_id`

andresriancho: This repeats in almost all endpoints, please fix in all.

w3af/core/ui/api/resources/sessions.py
```diff
@@ -0,0 +1,343 @@
...
+            opt_type = plugin_opts[opt_name].get_type()
+        except BaseFrameworkException:
+            return jsonify({
+                'code': '400',
+                'message': '%s is not a valid option for plugin %s' %
+                    (opt_name,
+                     plugin)
+                }), 400
+        try:
+            if opt_type.lower() == 'url_list':
```
andresriancho: I believe that this is not required, if you do a `plugin_opts[opt_name].set_value(['http://x.com', 'http://y.com'])` it should work.

w3af/core/ui/api/resources/sessions.py
```diff
@@ -0,0 +1,343 @@
...
+            else:
+                plugin_opts[opt_name].set_value(opt_value)
+        except (AttributeError, BaseFrameworkException):
+            return jsonify({
+                'code': '422',
+                'message': 'Invalid %s option %s' % (opt_type, opt_value)
+                }), 422
+
+    with SET_OPTIONS_LOCK:
+        if 'enabled' in request.json:
```
andresriancho: Didn't you handle the `enabled` parameter before?

w3af/core/ui/api/resources/sessions.py
```diff
@@ -0,0 +1,343 @@
...
+                                     plugin,
+                                     plugin_type,
+                                     enable=request.json['enabled'])
+            return jsonify({
+                'message': 'success',
+                'modified': request.json
+                })
+    try:
+        # Remove 'enabled' if included as we must set this separately
+        opt_names.pop('enabled')
```
andresriancho: We're going to have a bug here when a plugin (for some strange reason) adds a configuration parameter name called `enabled`.

Maybe you can add code to `OptionList` to make sure that never happens? I don't mind restricting the plugin developers like that.

Another solution is to have the `enabled` and plugin-related options in different JSON "namespaces".

w3af/core/ui/api/resources/sessions.py
```diff
@@ -0,0 +1,343 @@
...
+        try:
+            opt_value = request.json[opt_name]
+            opt_type = core_opts[opt_name].get_type()
+        except BaseFrameworkException:
+            return jsonify({
+                'code': '400',
+                'message': '%s is not a valid option here' % opt_name
+                }), 400
+        try:
+            if opt_type.lower() == 'url_list':
```
andresriancho: Same as above

w3af/core/ui/api/resources/sessions.py
```diff
@@ -0,0 +1,343 @@
...
+    w3af = SCANS[scan_id].w3af_core
+
+    if core_setting == 'http':
+        configurable = w3af.uri_opener.settings
+    if core_setting == 'misc':
+        configurable = MiscSettings()
+    if core_setting == 'target':
+        configurable = w3af.target
+
+    core_opts = configurable.get_options()
```
andresriancho: The next ~30 lines look very much like the configuration of plugin parameters. Maybe you should avoid code duplication and move this to a helper?

w3af/core/ui/api/tests/test_session_scan.py
```diff
@@ -0,0 +1,205 @@
...
+class APIScanFromSessionTest(IntegrationTest):
+
+    def test_start_scan_from_session(self):
+        plugin_settings = {
+            "crawl.web_spider": {
+                "only_forward": True,
+                "follow_regex": ".*"},
+            "audit.sqli": {}
+        }
+        target_url = "http://127.0.0.1:8000/audit/sql_injection/"
```
andresriancho: Use `get_moth_http` (or similar function, see other tests) to avoid hard-coding the port

w3af/core/ui/api/tests/test_session_scan.py
```diff
@@ -0,0 +1,205 @@
...
+                    self.api_url, scan_id, cs),
+                auth = self.api_auth,
+                data=json.dumps(core_settings[cs]),
+                headers=self.headers)
+
+        start_scan = requests.post(
+            '%s/sessions/%s/start' % (
+                self.api_url, scan_id),
+            auth = self.api_auth,
+            headers=self.headers)
```
andresriancho: Test has really good coverage, well done :+1:

w3af/core/ui/api/tests/test_sessions.py
```diff
@@ -0,0 +1,133 @@
...
+
+        session_id = json.loads(create_session.data)['id']
+        self.assertEqual(create_session.status_code, 201)
+
+        get_plugin_types = self.app.get('/sessions/%s/plugins/' % session_id,
+                                       headers=self.HEADERS)
+        plugin_types = json.loads(get_plugin_types.data)['entries']
+        self.assertEqual(get_plugin_types.status_code, 200)
+        self.assertIsInstance(plugin_types, list)
+        self.assertTrue('audit' in plugin_types)
```
andresriancho: `self.assertIn` should be used in these cases

w3af/core/ui/api/tests/test_sessions.py
```diff
@@ -0,0 +1,133 @@
...
+            '/sessions/%s/plugins/audit/this_should_also_404/' % session_id,
+            '/sessions/%s/core/nonexistent_setting/' % session_id
+            ]
+
+        for endpoint in test_404s:
+            should_404 = self.app.get(endpoint,
+                                       headers=self.HEADERS)
+            code = json.loads(should_404.data)['code']
+            self.assertEqual(code, 404)
+            self.assertEqual(should_404.status_code, 404)
```
andresriancho: This is a really long test that looks like it could be split in various tests. It's always better to have isolated tests which assert specific things.

andresriancho:
PR review finished! The code looks good, there are tests which PASS, everything seems to be complete except from :skull: :skull: :skull: documentation. Please consider adding documentation for the `sessions` endpoint to https://github.com/andresriancho/w3af/tree/master/doc/sphinx/api , follow the formatting/wording from the other endpoints. Documentation is critical since we want people to consume the API, and without docs they will find it really hard.

Add the docs to this PR, I'll review them, run the tests and code and finally merge.

w3af/core/ui/api/resources/sessions.py
OwenTuz: I receive an AttributeError if I do this - it looks like the code in set_value is expecting a URL object and calling verify_url:
`{
  "code": 500, 
  "exception_type": "AttributeError", 
  "filename": "target.py", 
  "function_name": "_verify_url", 
  "line_number": 101, 
  "message": "'unicode' object has no attribute 'get_protocol'", 
  "please": "https://github.com/andresriancho/w3af/issues/new"
}`

Presumably the user could call URL() and pass a URL object in that list with some success, but it seemed less friendly than plain text: there may be a better way to handle this?

w3af/core/ui/api/resources/sessions.py
OwenTuz: I've added a comment to clarify this:

`     if ('enabled' in opt_names and len(opt_names) == 1):
+    # If the request only contains 'enabled', set it here, return and skip
+    # the additional logic.
+    # If the request contains 'enabled' alongside other options, we handle it
+    # in the same lock section as other options below.
+    #
+    # This is done to ensure that the PATCH request is always atomic.
         with SET_OPTIONS_LOCK:
             enable_or_disable_plugin(w3af,`

I think it's valid, as the PATCH request must be atomic, but perhaps there's a neater way to handle it?

w3af/core/ui/api/resources/sessions.py
OwenTuz: Good catch, thanks. I'm afraid I've taken the more restrictive path and caught this in the 'add' method for `OptionList` - while the namespace method seems more elegant, it seems as though it would be confusing to allow 'enabled' as a plugin option.

OwenTuz:
Thanks for taking the time to review it, and for the feedback - I've implemented some of these and commented on others. 

Still to do is refactoring to reduce code reuse, so I'll add that as soon as I get a chance, along with any other fixes/edits based on the above.

One note - I'm finding that get_moth_http() returns 'fallback' and breaks the scan test, but it does so for existing tests as well. I'm getting to grips with a Mac for the first time due to work VPN requirements (Citrix >8( long story) and suspect this is my environment rather than the code.

If it does fail your CI tests, though, I apologise - let me know and I'll figure out an alternative.

OwenTuz:
Also, yes. The documentation :disappointed:

w3af/core/controllers/core_helpers/status.py
OwenTuz: This is called from `/scans/<scan_id>/status`, via `w3af.core.status.get_status_as_dict()`
The dict looks like this:
```
 data = {                                                        
         'status': self.get_simplified_status(),                 
         'is_paused': self.is_paused(),                          
         'is_running': self.is_running(),                        
                                                                 
         'active_plugin':                                        
             {'crawl': self.get_running_plugin('crawl'),         
              'audit': self.get_running_plugin('audit')}         
         ,                                                       
                                                                 
         'current_request':                                      
             {'crawl': crawl_fuzzable_request,                   
              'audit': audit_fuzzable_request},                  
                                                                 
         'queues':                                               
             {'crawl':                                           
                 {'input_speed': self.get_crawl_input_speed(),   
                  'output_speed': self.get_crawl_output_speed(), 
                  'length': self.get_crawl_qsize()},             
              'audit':                                           
                 {'input_speed': self.get_audit_input_speed(),   
                  'output_speed': self.get_audit_output_speed(), 
                  'length': self.get_audit_qsize()}              
             },                                                  
                                                                 
         'eta':                                                  
             {'crawl': self.get_crawl_eta(),                     
              'audit': self.get_audit_eta()},                    
                                                                 
         'rpm': self.get_rpm(),                                  
         }                                                       
```

Of those, only `get_rpm()` was throwing an exception if called before the scan was started - the rest return 'None', though functions such as `get_crawl_eta()` are equally meaningless if the scan has never run.

It may be more correct to patch the other functions so that they throw an exception, and simply handle the exceptions in `/scans/<scan_id>/status` instead?

OwenTuz:
Updated, with notes re some of your comments - there may well be better ways to do this, so please let me know if you have any preference.

andresriancho:
Going to work on this PR today, I'll test manually, review the source and merge. These changes are really valuable and should be in develop/master

PR #13005, opened 2015-11-09
Syntax Error - py3k compatibility

mattgia:
It didn't work in python 3.5

### DIFF
--- a/w3af_console
+++ b/w3af_console
@@ -62,7 +62,7 @@ def _configure_output_manager():
     """ 
     try:
         om.manager.set_output_plugins(['console'])
-    except Exception, e:
+    except Exception as e:
         print('Something went wrong, w3af failed to start the output manager.')
         print('Exception: "%s"' % e)
         sys.exit(-9)


andresriancho:
I'm not sure if accepting this makes sense, since once we change this line, w3af will crash in the next `py` that is not written according to py3k. In other words, what good is this for without a full py3k code migration?

This PR is very similar to https://github.com/andresriancho/w3af/pull/13006

matthew9012:
Sorry, I was a bit hasty with that change. I didn't realize the syntax was actually different between py2k and py3k. Also, I failed to read python 2.7 was a prerequisite. Sorry for wasting your time.

matthew9012:
If Python 2.7 is required, why not make the top line
#!/usr/bin/env python2.7
rather than
#!/usr/bin/env python

python is usually symbolically linked to the latest python version (3.5.0 in my case).

Just for people like me that don't read :P

andresriancho:
https://github.com/andresriancho/w3af/issues/13012

PR #13420, opened 2016-01-28
Silences faulty output plugin recommendation

niklasfemerstrand:
Only recommend to enable output plugins when none are enabled.

### DIFF
--- a/w3af/core/ui/console/rootMenu.py
+++ b/w3af/core/ui/console/rootMenu.py
@@ -79,7 +79,7 @@ class rootMenu(menu):
         """
         # Check if the console output plugin is enabled or not, and warn.
         output_plugins = self._w3af.plugins.get_enabled_plugins('output')
-        if 'console' not in output_plugins:
+        if 'console' not in output_plugins and len(output_plugins) == 0:
             msg = ("\nWarning: You disabled the console output plugin. If you"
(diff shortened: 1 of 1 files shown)


andresriancho:
Hmmm, not sure that I understand this change. Could you explain it?

qnrq:
When the output.console plugin is disabled you get the warning that says:

```python
            msg = ("\nWarning: You disabled the console output plugin. If you"
                   " start a new scan, the discovered vulnerabilities won\'t be"
                   " printed to the console, we advise you to enable at least"
                   " one output plugin in order to be able to actually see the"
                   " the scan output.")
```

You are given a recommendation to "enable at least one output plugin" even when there are other output plugins active. That's what the len check prevents: a recommendation to enable output plugins when there are output plugins active.

This change does the same len check as you already have in w3af/core/ui/console/plugins.py for the same message https://github.com/andresriancho/w3af/blob/master/w3af/core/ui/console/plugins.py#L195

So, the result of this fix is that there is no message recommending you to enable output plugins when there is more than 0 output plugins enabled. Currently the master branch will tell you to enable at least one output plugin if output.console is disabled, regardless of other output plugins being enabled.

andresriancho:
Thanks for explaining! :+1:

Issue #9635, opened 2015-04-25
Create plugin to detect insecure eval() calls in Node.js


andresriancho:
## Contributor wanted

This issue is [easy](https://github.com/andresriancho/w3af/issues?q=is%3Aopen+is%3Aissue+label%3Aeasy) and looking for a contributor, just add a comment saying you'll do it's yours!

Not sure where to start from? Take a look at our [First steps as a contributor](https://github.com/andresriancho/w3af/wiki/First-steps-as-a-contributor) and [Contributing 101](https://github.com/andresriancho/w3af/wiki/Contributing-101) wiki pages.

Other interesting tasks which can be completed by new `w3af` contributors can be found [here](https://github.com/andresriancho/w3af/issues?q=is%3Aopen+is%3Aissue+label%3Aeasy).
## Task description

Create a new audit plugin to detect insecure eval() calls in node.js as described here:
http://blog.gdssecurity.com/labs/2015/4/15/nodejs-server-side-javascript-injection-detection-exploitati.html

andresriancho:
@inkz contributed his time and knowledge and added code which allows w3af to detect (most I believe) of these types of vulnerabilities.

A question is still open, and I believe that @inkz is the best to answer: "Are we finding all the vulnerability cases explained in [this](http://blog.gdssecurity.com/labs/2015/4/15/nodejs-server-side-javascript-injection-detection-exploitati.html) blog post?"

Should we modify the plugin in some way to test more payloads? Something that I thought about while reading the blog post is that right now we're sending payloads like `Thread.Sleep(5);` but in the blog post they send `;Thread.Sleep(%s);`

Something else that we might be missing is a case like this (pseudo-code):

```
code = 'var foo = "'+ user_controlled_input + '"; run(foo);'
eval(code);
```

In cases like that we would have to send a payload like `";Thread.Sleep(5);var x="`, right?

What do you think @inkz ?

inkz:
Yes i think its better to add ";" to payload, should i hardcode it or there are other ways to do it?

andresriancho:
Well, it wouldn't be the just the `;` payload, it would be:
 * Send payload as specified in [these source code lines](https://github.com/andresriancho/w3af/blob/master/w3af/plugins/audit/eval.py#L46-L76)
 * Send payload as specified in [these source code lines](https://github.com/andresriancho/w3af/blob/master/w3af/plugins/audit/eval.py#L46-L76) prepended with `;`
[...]

PR #13436, opened 2016-02-03
fix for issue #9635

inkz:
added more payloads this time


inkz:
@andresriancho hi, cant find you in irc, i'm ready for a next task, if this request is fine

PR #13574, opened 2016-03-06
Enhanced XSS plugin

buglloc:
First of all, the new contexts in develop branch are awesome, great job! But they can be even better and I made several improvements (hopes it was useful).
#### Deal with the minor server-side payload modifications

Current implementation expects that the payload will be echoed untouched, but it is not always true, few real world examples:
- Payload in HTML-events are converted to HTML-safe sequences:

``` plain
url: http://w3af-buglloc.rhcloud.com/html-event-encoded-quot.php?input=input
input: 1234"1234
output: <a href="#" onclick='console.log("1234&quot;1234")'>click</a>
```
- Web application does not escape backslash in JavaScript string literals:

``` plain
url: http://w3af-buglloc.rhcloud.com/js-literal-backslash.php?input=input

input: 1234'1234
output:
<script>
var foo = '1234\'1234';
</script>


input: 1234\'1234
output:
<script>
var foo = '1234\\'1234';
</script>
```
- Web application escaped HTML-attribute as JavaScript string literals:

``` plain
url: http://w3af-buglloc.rhcloud.com/html-attr-escaped-js.php?input=input
input: 1234"1234
output: <div data-input="1234\"1234"></div>
```

In all of those cases w3af doesn't reports anything, because can't find payloads in the output. But it is still a potential XSS vulnerability, am i right?
I propose the following algorithm:
1. Send `left-random-boundPAYLOADright-random-bound` instead of `randomPAYLOADrandom`, this allows more accurate search for the modified payload
2. Do not search payload in output. Instead of this search `left-random-bound[anything]right-random-bound`, so if web application modified `' -> &#039;` we can deal with it
#### Context-guided scan for Reflected-XSS

Prepared set of vectors/payloads is OK while searching Stored-XSS, but we can improve checks for Reflected-XSS. We have all of contexts where payload echoed, so why we not used this useful information?
For example, JavaScript single line comment in [web application](http://w3af-buglloc.rhcloud.com/js-comment-sl-partial.php?input=input):

``` php
<?php
$value = strtr($_GET['input'], [
        "\n" => '\n',
        "\r" => '\r',
        "\\" => '\\\\',
        "'" => "\\'",
        '"' => '\\"',
        "</" => '\x3C\x2F'
    ]);
?>
<script>
// <?=$value?>

</script>
```

Let's look what XSS plugin [have for this case](https://github.com/andresriancho/w3af/blob/develop/w3af/plugins/audit/xss.py#L66):

``` python
    # Escape single line comments in JavaScript
    "\n",
```

Alright, I guess w3af does not report anything for this Reflected-XSS expect fp, because doesn't have vector with `\u2028`, `\u2029`. But if we rewrite XSS plugin to be context-guided we easily deal with this.
So I propose to change the algorithm as follows:
1. Send "empty" payload
2. Collect payloads from all of contexts where it was echoed
3. Test it:)
#### Minor improvements
- Highlights "text" tags (e.g. `title`, `textarea`, etc ) into separated class, which breaks `</` instead of `<`. It's safe from false positive like this:

``` plain
url: http://w3af-buglloc.rhcloud.com/js-literal-backslash.php?input=input
input: 1234<1234
output:
<script>
var foo = '1234<1234';
</script>


input: 1234</1234
output:
<script>
var foo = '1234\x3C/1234';
</script>
```
- Added `\u2028`, `\u2029` to JavaScript line terminators according to [ECMAScript specification](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-line-terminators)
- Added logic to handle HTML-like comments in JavaScript code according to [ECMAScript specification](http://www.ecma-international.org/ecma-262/6.0/index.html#sec-html-like-comments)
- JS/CSS string contexts now also checks escaping
#### Results

I wrote simple tests `http://w3af-buglloc.rhcloud.com/` for described cases, let's look to results!:)
In all tests i used `web_spider` as crawler and `xss` plugin w/o Stored-XSS search:
- w3af's `master` branch:

``` plain
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-comment.php", using HTTP method GET. The sent data was: "input=&type=sl" The modified parameter was "input". This vulnerability was found in the request with id 47.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/html-event-encoded-quot.php", using HTTP method GET. The sent data was: "input=cpq9d%60cpq9d" The modified parameter was "input". This vulnerability was found in the request with id 69.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-literal-backslash.php", using HTTP method GET. The sent data was: "input=2r1cv%2F%2A" The modified parameter was "input". This vulnerability was found in the request with id 98.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-comment-sl-partial.php", using HTTP method GET. The sent data was: "input=gyjna%2F%2A" The modified parameter was "input". This vulnerability was found in the request with id 106.

```
- w3af's `develop` branch:

``` plain
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-comment-sl-partial.php", using HTTP method GET. The sent data was: "input=7qadz%3C7qadz" The modified parameter was "input".This vulnerability was found in the request with id 52.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-comment.php", using HTTP method GET. The sent data was: "input=&type=ml" The modified parameter was "input".This vulnerability was found in the request with id 66.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-literal-backslash.php", using HTTP method GET. The sent data was: "input=o50yp%3Co50yp" The modified parameter was "input".This vulnerability was found in the request with id 133.

```
- w3af's with this PR:

``` plain
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-comment.php", using HTTP method GET. The sent data was: "...input=uva7l<tag`-->'</tag
*/</":"'*/:( =uva7r..." The modified parameter was "input".This vulnerability was found in the request with id 38.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-literal-backslash.php", using HTTP method GET. The sent data was: "input=m3fyl%5C%27m3fyr" The modified parameter was "input".This vulnerability was found in the request with id 44.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/html-event-encoded-quot.php", using HTTP method GET. The sent data was: "input=3tqvl%223tqvr" The modified parameter was "input".This vulnerability was found in the request with id 47.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/js-comment-sl-partial.php", using HTTP method GET. The sent data was: "input=in32l%E2%80%A9in32r" The modified parameter was "input".This vulnerability was found in the request with id 61.
A Cross Site Scripting vulnerability was found at: "http://w3af-buglloc.rhcloud.com/html-attr-escaped-js.php", using HTTP method GET. The sent data was: "input=7jhcl%227jhcr" The modified parameter was "input".This vulnerability was found in the request with id 72.

```


buglloc:
I know too many changes but to separate them without breaking, I could not :(
Also seen a similar idea on AppSecUSA 2015: [New Methods in Automated XSS Detection by Ken Belva](https://www.youtube.com/watch?v=G539NwvpL3I)

andresriancho:
> ## Deal with the minor server-side payload modifications
> In all of those cases w3af doesn't reports anything, because can't find payloads in the output. But it is still a potential XSS vulnerability, am i right?

I believe all those cases are NOT potential XSS because using the input from the examples you can NOT escape the contexts you show.

> I propose the following algorithm:
> * Send left-random-boundPAYLOADright-random-bound instead of randomPAYLOADrandom, this allows more accurate search for the modified payload
> Do not search payload in output. Instead of this search left-random-bound[anything]right-random-bound, so if web application modified ' -> &#039; we can deal with it

I believe that might actually improve our detection, yes.

> So I propose to change the algorithm as follows:
>Send "empty" payload
>Collect payloads from all of contexts where it was echoed
>Test it:)

I don't understand this. Will read code later.

w3af/core/data/context/context/html.py
```diff
@@ -133,12 +156,18 @@ class HTMLAttrQuoteGeneric(BaseContext):
     JS_PATTERN = re.compile('^ *javascript:', re.IGNORECASE)
     VB_PATTERN = re.compile('^ *vbscript:', re.IGNORECASE)
 
-    def __init__(self, payload, attr_name, attr_value):
+    # What about &NewLine;, &Tab;, etc?
+    JS_PROTOCOLS = {'javascript:', 'javascript&colon;',
```
andresriancho: Is this actually supported by any browser? I mean... will any browser run:

```html
<a href="javascript&colon;foo(2)">
```

w3af/core/data/context/context/javascript.py
```diff
@@ -23,12 +23,16 @@
...
 
 STRING_DELIMITERS = {'"', "'"}
+# According to http://www.ecma-international.org/ecma-262/6.0/index.html
+#                                                   #sec-line-terminators
+LINE_TERMINATORS = {'\n', '\r', u'\u2028', u'\u2029'}
 
 
 class ScriptSingleLineComment(BaseContext):
-    CAN_BREAK = {'\n', '\r'}
+    CAN_BREAK = LINE_TERMINATORS
```
andresriancho: If we make this change we should also add a payload that sends `\u2028` and `\u2029` in xss.py

w3af/core/data/context/context/javascript.py
```diff
@@ -36,7 +40,25 @@ class ScriptMultiLineComment(BaseContext):
...
+                escaped = not escaped
+            elif c == self.ATTR_DELIMITER and not escaped:
+                return True
+            else:
+                escaped = False
+
+        return False
+
+    def get_payloads(self):
+        return {self.ATTR_DELIMITER, '\\%s' % self.ATTR_DELIMITER}
```
andresriancho: If the input is `xxx"yyy` and the context is `<a href="xxx\"yyy">...` then we don't have a vulnerability since we won't be able to escape the attribute value context, correct? Then, why handling all those escapes?

w3af/core/data/context/context/main.py
```diff
@@ -78,16 +82,21 @@ def get_context_iter(data, payload):
...
 class ContextDetectorHTMLParser(HTMLParser):
 
-    def __init__(self, payload):
+    RAW_TEXT_TAG = {
+        'title',
+        'textarea',
+        'plaintext',
+        'xmp',
+        'listing'
+    }
```
andresriancho: Why are these tags special?

w3af/plugins/audit/xss.py
```diff
@@ -134,18 +146,38 @@ def _report_vuln(self, mutant, response, mod_value):
...
-        payload = replace_randomize(''.join(self.PAYLOADS))
+        payload = place_boundary(''.join(payloads))
 
         trivial_mutant = mutant.copy()
         trivial_mutant.set_token_value(payload)
-        
+
         response = self._uri_opener.send_mutant(trivial_mutant)
 
         # Add data for the persistent xss checking
```
andresriancho: We need randomized payloads here to identify source->sink in persistent xss across multiple pages.

andresriancho:
The PR is too large and at least to me it tries to fix something that's not broken.

PHP test files are missing from the PR, so I can't re-test in my environment. I recommend replacing all those PHP helper files with unittests (like the ones we already have for the context feature).

Before doing anything else please explain clearly which problem the PR wants to fix.

buglloc:
Thanks for taking the time to look!

> Before doing anything else please explain clearly which problem the PR wants to fix.

This PR tries to improve XSS auditor by implement two ideas - fuzzy payload search and increase number of payloads depending on context for Reflected-XSS instead of using prepared payload sets. And I believe this changes helps in the future for another not trivial XSS cases. Current plugin not broken, but can be better.

> I believe all those cases are NOT potential XSS because using the input from the examples you can NOT escape the contexts you show.

Why do you think so?:( All of my examples can break their contexts and I saw those cases in the real world applications.
For example web-application converts quotes to html entities (e.g. `"` -> `&#34;` or `&#x22;` or `&quot;`) and echoed `input` in the JS string literal placed in HTML-event: 
```plain
input: 12341234
output: <a href="#" onclick='console.log("12341234")'>click</a>
```
And if we send `1234"1234` application will output:
```html
<a href="#" onclick='console.log("1234&quot;1234")'>click</a>
```
Original payload `1234"1234` doesn't echoed in the output, but after server-side modifications  (i.e. `1234&quot;1234`) payload still can breaks the JavaScript double encoded string in HTML-event, because browser will unescape all html entities before execution and tries to execute `console.log("1234"1234")`.
This PR can catch whose potential XSS vulnerabilities and do so.

> PHP test files are missing from the PR, so I can't re-test in my environment. I recommend replacing all those PHP helper files with unittests (like the ones we already have for the context feature).

I have already did this, [one of those tests](https://github.com/buglloc/w3af/commit/2301d4f86fab5e6d320cc0676be156e78ef42736#diff-835fced60b704d944bf2a19f06d0440bR277). PHP test files used in description for illustration purposes only, perhaps it did not work as i expected:( If this could helps - I [published it](https://github.com/buglloc/w3af-pr13574-examples).


>>  So I propose to change the algorithm as follows:
>> Send "empty" payload
>> Collect payloads from all of contexts where it was echoed
>> Test it:)

> I don't understand this. Will read code later.

I'm sorry, maybe example could help...
Tested script: [https://github.com/buglloc/w3af-pr13574-examples/blob/master/multiple-contexts.php](https://github.com/buglloc/w3af-pr13574-examples/blob/master/multiple-contexts.php)
Payload was echoed in 4 types of contexts `HtmlText`, `ScriptText`, `ScriptSingleLineComment` and
 `ScriptSingleQuoteString`
XSS plugin w/o persistent XSS checks
Explained w3af log:
```plain
Starting plugin: audit.xss
Created 1 mutants for "Method: GET | http://w3af-buglloc.rhcloud.com/multiple-contexts.php | Query string: (input)" (query string: 1)
// Collect payloads from all of contexts where it was echoed
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=onezlonezr returned HTTP code "200" (id=36,from_cache=0,grep=1)

// Identify trivial XSS
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=jlajl<tag`-->'
\'*/</tag</":"'*/:( =jlajr returned HTTP code "200" (id=37,from_cache=0,grep=1)

// Test each payload
Created 8 mutants for "Method: GET | http://w3af-buglloc.rhcloud.com/multiple-contexts.php | Query string: (input)" (query string: 8)

// u'\u2029' from ScriptSingleLineComment
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=ykkhlykkhr returned HTTP code "200" (id=38,from_cache=0,grep=1)

// '<' from HtmlText
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=wcgql<tagwcgqr returned HTTP code "200" (id=39,from_cache=0,grep=1)

// ' from ScriptSingleQuoteString
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=qof7l'qof7r returned HTTP code "200" (id=40,from_cache=0,grep=1)

// '\n' from ScriptSingleLineComment
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=mx1ll
mx1lr returned HTTP code "200" (id=42,from_cache=0,grep=1)

// '</' from ScriptText
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=yrail</yrair returned HTTP code "200" (id=43,from_cache=0,grep=1)

// '\r' from ScriptSingleLineComment
h9obr returned HTTP code "200" (id=41,from_cache=0,grep=1)input=h9obl

// u'\u2028' from ScriptSingleLineComment
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=eptbleptbr returned HTTP code "200" (id=44,from_cache=0,grep=1)

// '\r' from ScriptSingleQuoteString
GET http://w3af-buglloc.rhcloud.com/multiple-contexts.php?input=lygkl\'lygkr returned HTTP code "200" (id=45,from_cache=0,grep=1)
```
As you can see this PR allows XSS plugin dynamically construct payload set, checks only it (for Reflected-XSS only) and test rare cases (e.g. `\u2028` for ScriptSingleLineComment)

buglloc:
I known the PR is too large :( I could try to divide it into a few smaller, but then each of them may not make sense. 

In any case, I am ready to do everything :)

w3af/core/data/context/context/html.py
buglloc: Sure! Why not? I'll try to find some research for PoC and link it in comment, may be on Shazzer

w3af/core/data/context/context/javascript.py
buglloc: It is not necessary due to changes in the PR case with `\u2028` and `\u2029` can leave only for Reflected-XSS and doesn't send it each time.
But if you think we need to check this case for persistent XSS - I'll add it.

w3af/core/data/context/context/javascript.py
buglloc: This escaping handled only for JavaScript string literals, your example about `HtmlAttrDoubleQuote` that doesn't apply to this case at all.
Hmm, for input `xxx"yyy` and the context `<a href="xxx\"yyy">...` we have a vulnerability since `"` can break `HtmlAttrDoubleQuote`.

w3af/core/data/context/context/main.py
buglloc: Because It handles raw content (`RCDATA`, `RAWTEXT`, etc) and not parsed nested HTML-tags, expect end tag.
In other words:
```html
This is NOT a XSS vulnerability
<title><svg onload=prompt`XSS`></title>

This is it
<title></title><svg onload=prompt`XSS`></title>
```
So for these tags I created new context `HtmlRawText` which can be breaks by `</`

buglloc: May be i must leave only `title` and `textarea` because `plaintext`, `xmp` and `listing` are deprecated and can not be interpreted by some browsers.
I'll will research this and make the changes if needed.

w3af/core/data/context/context/html.py
andresriancho: > Why not?

Browser expects `<a href="javascript:foo(2)">` not `<a href="javascript&colon;foo(2)">`

w3af/core/data/context/context/javascript.py
andresriancho: Ah! Tested it and it works as you said. I expected browsers to use `\"` as escape of the `"`

My test was with chrome and `<a x="xxx\" href=\"/foobar\" yyy">test</a>`

I still don't understand why you need to handle those escapes. The old version of the code should also support them. By support I mean that it was possible to find XSS? We should chat over GTalk so you can explain these changes to me.

andresriancho:
> increase number of payloads depending on context for Reflected-XSS instead of using prepared payload sets

If the logic behind this is:
 * Send boundary and extract context type/name based on where it gets echoed.
 * Send payloads only to break the previously identified context

That strategy is really good to reduce the number of HTTP requests being sent; but might fail to detect XSS in error handling functions:

```
if is_alpha_num(user_input):
    echo "<a href='%s'>test</a>" % user_input
else:
    # Enter here when the input is not alphanumeric
    echo "Invalid input %s" % user_input
```

Or:

```
if is_alpha_num(user_input):
    echo "Thanks!"
else:
    # Enter here when the input is not alphanumeric
    echo "Invalid input %s" % user_input
```

The implemented strategy will send [all payloads](https://github.com/andresriancho/w3af/blob/develop/w3af/plugins/audit/xss.py#L178-L187) which requires more bandwidth but has more coverage. I believe we should either keep the current strategy or give the user the option to choose between the two.

w3af/core/data/context/context/html.py
buglloc: > Browser expects <a href="javascript:foo(2)"> not <a href="javascript&colon;foo(2)">

Yes, but browser will unescape all html entities before processing uri. So `javascript&colon;foo(2)` transforms to `javascript:foo(2)` and successfully works. That is why some WAFs can be bypassed by using html5 entities, e.g. `<a href="javascript&Tab;&colon;foo(2)">`

andresriancho:
The fuzzy payload search, now that I understand the cases with `\"` and `onclick='console.log("1234&quot;1234")'>click</a>` are actually a really good improvement to w3af's XSS detection!

buglloc:
> The implemented strategy will send all payloads which requires more bandwidth but has more coverage. I believe we should either keep the current strategy or give the user the option to choose between the two.

You're absolutely right! Maybe add the option as you say is a good choice. Or just simply continue to send all known payloads and use contexts for coverage specific cases, which is not reasonable to send each time (e.g. `\u2028`, `\u2029` for single line comment)?

qdii:
Any news on this PR? this looks very promising

PR #13647, opened 2016-03-20
New CSP module and grep plugin

oxdef:
1. Re-factored CSP module: from functions to OOP (directive->policy->holder of policies) Now much easier to maintain the code and add new functionality 
2. Added retrieval of CSP policy from meta tag
3. Added new checks
4. Removed related to not standard  CSP realization. Now only modern CSPv2
5. Updated tests

### DIFF
--- a/w3af/core/controllers/csp/tests/test_csp_utils.py
+++ b/w3af/core/controllers/csp/tests/test_csp_utils.py
@@ -346,7 +346,7 @@ class TestUtils(unittest.TestCase):
         """
         Test case in which we add untrusted hosts into somes policies.
         """  
-        header_value = "default-src 'self'; script-src 'self' trust.com evil.com;"
+        header_value = "default-src 'self'; script-src blob: data: 'self' trust.com evil.com;"
         csp = CSPPolicy()
         csp.trusted_hosts = ['trust.com']
         csp.init_value(header_value)
--- a/w3af/core/controllers/csp/tests/test_csp_utils.py
+++ b/w3af/core/controllers/csp/tests/test_csp_utils.py
@@ -173,7 +173,7 @@ class TestUtils(unittest.TestCase):
                        " object-src media1.example.com media2.example.com"\
                        " *.cdn.example.com; script-src trustedscripts.example.com;"\
                        " form-action foo.com/ctxroot/action1 foo.com/ctxroot/action2;"\
-                       " plugin-types application/pdf audio/ogg;"
+                       " plugin-types application/pdf;"
         csp = CSPPolicy()
         csp.init_value(header_value)
         
@@ -197,9 +197,8 @@ class TestUtils(unittest.TestCase):
         self.assertTrue("foo.com/ctxroot/action1" in d.source_list)
         self.assertTrue("foo.com/ctxroot/action2" in d.source_list)
         d = csp.get_directive_by_name("plugin-types")
-        self.assertEqual(len(d.media_type_list), 2)
+        self.assertEqual(len(d.media_type_list), 1)
         self.assertTrue("application/pdf" in d.media_type_list)
-        self.assertTrue("audio/ogg" in d.media_type_list)
 
     def test_retrieve_csp_policies_with_special_policies_case02(self):
         """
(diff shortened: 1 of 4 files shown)


oxdef:
According to the log of CircleCI I don't understand how it is connected with this pull request :(

andresriancho:
@oxdef shouldn't be your problem. I believe builds have been broken for some time now. CircleCI changes things, pip modules get outdated, etc.

w3af/core/controllers/csp/tests/test_csp_utils.py
```diff
@@ -173,27 +109,17 @@ def test_provides_csp_features_no_case02(self):
...
-        #     img-src -> image-src
         header_value = "default-src ' '; img-src ' '"
-        hrds = {CSP_HEADER_W3C: header_value}.items()
-        csp_headers = Headers(hrds)
-        
-        http_response = HTTPResponse(200, '', csp_headers, self.url, self.url)
-                
-        self.assertFalse(provides_csp_features(http_response))
+        csp = CSPPolicy()
+        print csp.init_value(header_value), csp.directives
```
andresriancho: Remove print

w3af/core/controllers/csp/tests/test_csp_utils.py
```diff
@@ -220,352 +142,119 @@ def test_provides_csp_features_yes_case02(self):
...
-        vulns = find_vulns(http_response)  
+        header_value = "default-src 'self';script-src 'self'; object-src 'self';" \
+                       "style-src 'self'; img-src 'self'; media-src 'self';" \
+                       "frame-src 'self'; font-src 'self'; sandbox;" \
+                       "form-action foo.com/myCtx/act; connect-src 'self';"\
+                       "plugin-types application/pdf;"
+        csp = CSPPolicy()
+        csp.init_value(header_value)
+        vulns = csp.find_vulns()
         self.assertEqual(len(vulns), 0)
```
andresriancho: I like these new tests, they are easier to read than the previous version.

w3af/plugins/grep/csp.py
```diff
@@ -19,158 +19,163 @@
...
-                for csp_vuln in csp_vulns_list:        
-                    if csp_vuln.desc.strip().lower() == ref:
-                        if vuln_store_item.resp_id not in list_resp_id:
-                            list_resp_id.append(vuln_store_item.resp_id)
+        :return: A list of option objects for this plugin.
+        """
+        ol = OptionList()
 
-        return list_resp_id
+        d = 'Maximum number of sources in source list'
```
andresriancho: This description doesn't make much sense if you don't read the code.

 * Change it to something meaningful
 * Add a help

w3af/plugins/grep/csp.py
```diff
@@ -19,158 +19,163 @@
...
+        o = opt_factory('source_limit', self._source_limit, d, 'integer')
+        ol.add(o)
 
+        d = 'Trusted hosts for finding untrusted ones in source list'
```
andresriancho: Same as above

w3af/plugins/grep/csp.py
```diff
@@ -19,158 +19,163 @@
...
+
+    def get_long_desc(self):
         """
-        Implements method from base class.
+        :return: A DETAILED description of the plugin functions and features.
         """
-        return ['url','resp_id']
\ No newline at end of file
+        return """
+        This plugin identifies incorrect or too permissive Content Security Policy version 2
```
andresriancho: Is version 2 back compatible with CSP version 1? Maybe the description should say "too permissive Content Security Policy versions 1 and 2" ?

w3af/plugins/grep/csp.py
```diff
@@ -19,158 +19,163 @@
...
+"""
+
+    vuln_without_tpl = """
+There were found URL without CSP. Please, review it:
+
+    %s
+"""
+
+    vuln_no_csp_tpl = """
+    The whole target has no CSP protection.
```
andresriancho: Unnecessary 4-spaces inside string?

oxdef:
Andres, thanks for feedback! All has been fixed except:

> Is version 2 back compatible with CSP version 1? Maybe the description should say "too permissive Content Security Policy versions 1 and 2" ?

[It looks like](
https://www.w3.org/TR/CSP/#changes-from-level-1
) CSP2 is incompatible with Level 1. As I know [1.1](https://www.w3.org/TR/2014/WD-CSP11-20140211/) also contains directives which 2 doesn't contain.

Also added 2 more options for the plugin:

1. To report when policy doesn't contain report-uri
2. Manage if you want to report unsafe-eval

oxdef:
Andres, do you have any other questions of feedback?

PR #13814, opened 2016-05-03
Update installation command for Fedora.

bheesham:
Fedora now uses `dnf` instead of the `yum` command to install packages.

---

This change is [Reviewable](https://reviewable.io/reviews/andresriancho/w3af/13814)


andresriancho:
This change is invalid, since it will break support for old fedora versions.

You should handle it in a different way, see how different versions of debian/ubuntu are handled-

Issue #8461, opened 2015-02-23
Allow w3af_console_docker parameter forwarding


andresriancho:
## Description
When the user runs `./w3af_console_docker` it would be awesome to be able to forward command line parameters to the container command. When this is done the following command:

```
./w3af_console_docker -s ~/w3af-shared/foo.w3af
```

Should output the same as:

```
./w3af_console -s ~/w3af-shared/foo.w3af
```

## Source
https://github.com/andresriancho/w3af/tree/master/extras/docker/

PR #14022, opened 2016-07-01
New Dockerfile and docker build scripts

andersonDadario:
Changes:

1) Fixes Issue [8461](https://github.com/andresriancho/w3af/issues/8461) -- at least through bash command line
2) W3af now runs using the user "app" instead of "root" inside docker container
3) The new Dockerfile exposes w3af_console, w3af_api and w3af_gui

> When the user runs ./w3af_console_docker it would be awesome to be able to forward command line parameters to the container command. When this is done the following command:  
> ./w3af_console_docker -s ~/w3af-shared/foo.w3af  
> Should output the same as:  
> ./w3af_console -s ~/w3af-shared/foo.w3af

It's now possible to call w3af_console like this:

`docker run --name w3af --rm -it -v /root/:/tmp andres/w3af w3af_console -s /tmp/script.w3af`

P.S.: Supposing that your `script.w3af` is located at `/root/script.w3af` in the host machine.
# 

More on interacting with W3af docker container:

Access container's bash: `docker exec -it w3af /bin/bash`
Check if w3af is running: `docker top w3af`
View w3af output: `docker logs w3af`
Get w3af's info (including IP Address): `docker inspect w3af`
Get w3af's IP address alone: `docker inspect -f '{{ .NetworkSettings.IPAddress }}' w3af`
# 

Note: W3af image is around 750 MB in size. It's ok I suppose. OpenVAS is 3.5GB and Owasp ZAP 1.5GB. W3af size is good then :)

Note: I've followed [Contributing 101](https://github.com/andresriancho/w3af/wiki/Contributing-101) but the pull request to master wasn't going to work as the "git branch" command copies the "origin/development" so here I am, making a pull request to "origin/development" :)
# 

What's missing (that I can tell so far):
- Didn't check if EXPOSE matches all necessary ports for w3af to run
- Didn't check how to run W3AF GUI (forward X), but forward X as a port may be an EXPOSE entry too
- Didn't check if is necessary to modify any other python script. I focused on Dockerfile alone
- I forgot to add Andres as the maintainer, but can add. No problem :)

### DIFF
--- a/extras/docker/Dockerfile
+++ b/extras/docker/Dockerfile
@@ -1,92 +1,21 @@
-# w3af.org
-# https://github.com/andresriancho/w3af/tree/master/extras
+FROM ubuntu:14.04
 
-FROM ubuntu:12.04
-MAINTAINER Andres Riancho <andres.riancho@gmail.com>
+ADD . /w3af
+RUN apt-get update && apt-get upgrade -y && \
+    /bin/bash /w3af/add_user.sh && \
+    /bin/bash /w3af/install.sh && \
+    /bin/bash /w3af/cleanup.sh
 
-# Initial setup
-RUN mkdir /home/w3af
-WORKDIR /home/w3af
-ENV HOME /home/w3af
-ENV LANG en_US.UTF-8
-ENV LANGUAGE en_US:en
-ENV LOGNAME w3af
-# Squash errors about "Falling back to ..." during package installation
-ENV TERM linux
-RUN echo 'debconf debconf/frontend select Noninteractive' | debconf-set-selections
+USER app
+ENV PATH /home/app/w3af:$PATH 
+WORKDIR /home/app
 
-# Update before installing any package
-RUN apt-get update -y
-RUN apt-get upgrade -y
-RUN apt-get dist-upgrade -y
+CMD["w3af_api","w3af_console","w3af_gui"]
 
-# Install basic and GUI requirements, python-lxml because it doesn't compile correctly from pip
-RUN apt-get install -y python-pip build-essential libxslt1-dev libxml2-dev libsqlite3-dev \
-                       libyaml-dev openssh-server python-dev git python-lxml wget libssl-dev \
-                       xdot python-gtk2 python-gtksourceview2 ubuntu-artwork dmz-cursor-theme \
-                       ca-certificates libffi-dev
+EXPOSE 5000,44449,44444
 
-# Add the w3af user
-# TODO - actually use the w3af user instead of running everything as root
-RUN useradd w3af
-
-# Get ssh package ready
-RUN mkdir /var/run/sshd
-RUN echo 'root:w3af' | chpasswd
-RUN mkdir /home/w3af/.ssh/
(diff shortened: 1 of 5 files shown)


extras/docker/Dockerfile
```diff
@@ -1,92 +1,21 @@
-# w3af.org
-# https://github.com/andresriancho/w3af/tree/master/extras
```
andresriancho: re-add these two lines.

extras/docker/Dockerfile
```diff
@@ -1,92 +1,21 @@
...
+FROM ubuntu:14.04
 
-FROM ubuntu:12.04
-MAINTAINER Andres Riancho <andres.riancho@gmail.com>
```
andresriancho: Why remove me?

extras/docker/Dockerfile
```diff
@@ -1,92 +1,21 @@
...
+ADD . /w3af
```
andresriancho: This is a bad idea. If you `ADD` a directory to the docker image, the cache will break and all the commands after it will be re-run instead of restored from cache. `ADD` commands should be run as late as possible to speedup builds.

andresriancho: Oh... I see... you broke the Dockerfile down into smaller scripts. Hmmm...

andresriancho:
> 1) Fixes Issue 8461 -- at least through bash command line
> ...
> docker run --name w3af --rm -it -v /root/:/tmp andres/w3af w3af_console -s /tmp/script.w3af
> ...
> P.S.: Supposing that your script.w3af is located at /root/script.w3af in the host machine.

This was possible before too I believe, so issue 8461 is not fixed?

andresriancho:
Maybe you should create a different PR to improve http://docs.w3af.org/en/latest/docker.html with these?

> More on interacting with W3af docker container:
> 
> Access container's bash: docker exec -it w3af /bin/bash
> Check if w3af is running: docker top w3af
> View w3af output: docker logs w3af
> Get w3af's info (including IP Address): docker inspect w3af
> Get w3af's IP address alone: docker inspect -f '{{ .NetworkSettings.IPAddress }}' w3af

andresriancho:
> Note: W3af image is around 750 MB in size. It's ok I suppose. OpenVAS is 3.5GB and Owasp ZAP 1.5GB. W3af size is good then :)

Good to know :)

extras/docker/Dockerfile
andersonDadario: I didn't, although it looks like it.

I built this Dockerfile from scratch instead of changing the existent, so the differences started to appear. The same applies for the lines above ("# w3af.org",etc )

Will add those lines, don't worry.

andersonDadario:
1) And yes, my bad, this PR doesn't fix #8461 -- It's all based on SSH and will need to be updated or removed I believe.

2) Regarding the PR http://docs.w3af.org/en/latest/docker.html I believe that is the right place, but most people will go directly to docker hub first. That's what happened to me, so finding such info the w3af readme there would be even better IMHO.

3) Requested lines added!

extras/docker/Dockerfile
andersonDadario: I may be wrong, but I believe this way works best. It's not an original idea of mine I confess. I saw such organization and as you can see, smaller scripts are easier to maintain than a big script.

PR #14188, opened 2016-08-12
fc24 packagename is python2-setuptools.noarch

sboschman:


peter-manton:
This is also the case on FC25

Issue #8891, opened 2015-03-14
Improvements for HTML report template


andresriancho:
The report template I've drafted is properly rendered but, as usual, ugly. These are some of the things I would like others to improve.

## Header

![top](https://cloud.githubusercontent.com/assets/865200/6651463/de233736-ca21-11e4-9093-cadd02f9ec69.png)

![top2](https://cloud.githubusercontent.com/assets/865200/6651465/17fdb6ac-ca22-11e4-97be-6a0560f61d61.png)

 - [ ] There should be some padding on the top, before the first div with data
 - [ ] More padding after the main title
 - [ ] Potentially use different font for titles
 - [ ] Solve the issue we see in the second screenshot where a long list of enabled plugins uses a lot of vertical space and leaves the left "blank". Maybe add one of those "More ..." and "Less ..." buttons that show/hide data?

## Vulnerability

![vuln](https://cloud.githubusercontent.com/assets/865200/6651469/51a2b3e4-ca22-11e4-8fa0-767576e6561b.png)
![vuln2](https://cloud.githubusercontent.com/assets/865200/6651482/c1eef4dc-ca22-11e4-8804-6f6a3f0c68d4.png)

 - [ ] The red flag icon is not properly located, it should be centered (vertically/horizontally) and have padding
 - [ ] Maybe change the icon with something better/nicer
 - [ ] Maybe make the icon smaller
 - [ ] Use highlight.js to highlight the HTTP request/response
 - [ ] When there are long HTTP responses we're showing them all, using a lot of space. It might be a good idea to have a "vertical limit" for the HTTP request/response and add the "More ..." and "Less ..." links (as explained above)

## Overall
 - [ ] Change the background so that the divs with data (which will still have white background) grab the reader's attention
 - [ ] Apply cosmetic changes to make the report look beautiful

Mruocco:
I have done a lot of this out of personal necessity. 
I will clean it up and issue a PR.

andresriancho:
@Mruocco sounds good! Please do!

preetpdsilva-netizen:
hey @andresriancho  i would like to help you in this issue can you assign to me .

PR #14216, opened 2016-08-19
Template cosmetic changes

AverageS:
Added spoilers to http response in proofs, added spoiler to included plugins, Added title "Results" before the actual results tab. In my opinion this is the fix for https://github.com/andresriancho/w3af/issues/8891

### DIFF
--- a/w3af/plugins/output/html_file/templates/complete.html
+++ b/w3af/plugins/output/html_file/templates/complete.html
@@ -22,6 +22,25 @@
           overflow: hidden;
           text-overflow: ellipsis;
         }
+        summary::-webkit-details-marker {
+          display: none
+        }
+        summary:after {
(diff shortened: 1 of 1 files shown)


andresriancho:
Just tested the modifications and I believe that while they improve the previous HTML report (where all the HTTP response was shown, which is inconveniente in some cases) it is not the ideal solution.

https://css-tricks.com/text-fade-read-more/ is what I would like to see implemented instead. It has these interesting features:
 * All CSS
 * We can specify a max-height for the HTTP request and response areas, CSS will add the read-more button
 * We don't need to make any changes to the python code (other read-more css tricks that do not use max-height required this)

I know... I know... I'm answering to a PR with a 3 year delay... if you're not interested in implementing this just let me know and I'll try to do it.

PR #14592, opened 2016-11-07
Fix Fedora Dependency Errors

felipenovais:
## This commit fix some dependency errors on Fedora Linux:

1. When checking for python-setuptools the actual package installed on fedora is not python-setuptools, actually when you run the command `yum/dnf install python-setuptools` the real package name is on rpm is python2-setuptools.

2. Another error is during pybloomfiltermmap compilation process because it requires the gcc-c++ package.


andresriancho:
Better late than never https://github.com/andresriancho/w3af/commit/271adc1275ee9ae216e29e7a10a75d5f86e4ac65

PR #14784, opened 2016-12-13
dont check signatures if the response is not 200 in crawl/find_backdoors

majormoses:
When tomcat (or most web servers I have worked with) returns a 404 it includes information from the request, as such this does not make a very interesting check if its a 404.

image

Origonally I decided to use a check that already exists to check if its a 404 and if so returns. but after thinking about it we should catch only 200s.

### DIFF
--- a/w3af/plugins/crawl/find_backdoors.py
+++ b/w3af/plugins/crawl/find_backdoors.py
@@ -118,27 +118,30 @@ class find_backdoors(CrawlPlugin):
         except BaseFrameworkException:
             om.out.debug('Failed to GET webshell:' + web_shell_url)
         else:
-            signature = self._match_signature(response)
-            if signature is None:
+            if response.get_code() == 200:
+                signature = self._match_signature(response)
+                if signature is None:
+                    return
+
+                desc = (u'An HTTP response matching the web backdoor signature'
+                        u' "%s" was found at: "%s"; this could indicate that the'
+                        u' server has been compromised.')
+                desc %= (signature, response.get_url())
+
+                # It's probability is higher if we found a long signature
+                _severity = severity.HIGH if len(signature) > 8 else severity.MEDIUM
+
+                v = Vuln(u'Potential web backdoor', desc, _severity,
+                         response.id, self.get_name())
+                v.set_url(response.get_url())
+
+                kb.kb.append(self, 'backdoors', v)
+                om.out.vulnerability(v.get_desc(), severity=v.get_severity())
+
+                fr = FuzzableRequest.from_http_response(response)
+                self.output_queue.put(fr)
+            else:
                 return
-
-            desc = (u'An HTTP response matching the web backdoor signature'
-                    u' "%s" was found at: "%s"; this could indicate that the'
-                    u' server has been compromised.')
-            desc %= (signature, response.get_url())
-
-            # It's probability is higher if we found a long signature
-            _severity = severity.HIGH if len(signature) > 8 else severity.MEDIUM
-
-            v = Vuln(u'Potential web backdoor', desc, _severity,
-                     response.id, self.get_name())
-            v.set_url(response.get_url())
-
-            kb.kb.append(self, 'backdoors', v)
-            om.out.vulnerability(v.get_desc(), severity=v.get_severity())
-
-            fr = FuzzableRequest.from_http_response(response)
-            self.output_queue.put(fr)
 
     def _match_signature(self, response):
         """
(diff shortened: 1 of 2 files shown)


majormoses:
Reason for the second commit is that it was working until I committed and then ran again. I saw this error message:
```
$ ./w3af_gui 
Starting w3af, running on:
  File "./w3af_gui", line 110, in <module>
    _main()
  File "./w3af_gui", line 106, in _main
    sys.exit(main())
  File "./w3af_gui", line 100, in main
    from w3af.core.ui.gui.main import main as gui_main
  File "/home/babrams/projects/personal/w3af/w3af/core/ui/gui/main.py", line 69, in <module>
    print get_versions()
  File "/home/babrams/projects/personal/w3af/w3af/core/controllers/exception_handling/helpers.py", line 88, in get_versions
    w3af_version = '\n    '.join(get_w3af_version().split('\n'))
  File "/home/babrams/projects/personal/w3af/w3af/core/controllers/misc/get_w3af_version.py", line 58, in get_w3af_version
    version_dict = get_w3af_version_as_dict()
  File "/home/babrams/projects/personal/w3af/w3af/core/controllers/misc/get_w3af_version.py", line 44, in get_w3af_version_as_dict
    cdate = ' - %s' % get_latest_commit_date() if is_git_repo() else ''
  File "/home/babrams/projects/personal/w3af/w3af/core/controllers/auto_update/utils.py", line 82, in get_latest_commit_date
    cdate = git.Repo(path).head.commit.committed_date
  File "/usr/local/lib/python2.7/dist-packages/gitdb/util.py", line 237, in __getattr__
    self._set_cache_(attr)
  File "/usr/local/lib/python2.7/dist-packages/git/objects/commit.py", line 132, in _set_cache_
    self._deserialize(StringIO(stream.read()))
  File "/usr/local/lib/python2.7/dist-packages/git/objects/commit.py", line 443, in _deserialize
    self.author.name = self.author.name.decode(self.encoding) 
LookupError: unknown encoding: -----BEGIN PGP SIGNATURE-----
```

After doing some research it appeared that the functionality did not show up until 0.3.2 rather than bumping it to the minimum requirement I thought it might be better to go to latest if we dont see issues.

w3af/plugins/crawl/find_backdoors.py
```diff
@@ -118,27 +119,30 @@ def _check_if_exists(self, web_shell_url):
         except BaseFrameworkException:
             om.out.debug('Failed to GET webshell:' + web_shell_url)
         else:
-            signature = self._match_signature(response)
-            if signature is None:
-                return
+            if not is_404(response):
```
majormoses: I was thinking that rather than checking to make sure its not a 404 it might be better to check if it is 200:
```python
if status == 200:
```

Any thoughts? I prefer the 200 approach personally/

majormoses:
@andresriancho can we get this and reviewed and merged?

andresriancho (changes requested):
Any comments on the question about false positives vs. performance?

w3af/plugins/crawl/find_backdoors.py
```diff
@@ -118,27 +118,30 @@ def _check_if_exists(self, web_shell_url):
         except BaseFrameworkException:
             om.out.debug('Failed to GET webshell:' + web_shell_url)
         else:
-            signature = self._match_signature(response)
-            if signature is None:
-                return
+            if response.get_code() == 200:
+                signature = self._match_signature(response)
+                if signature is None:
+                    return
```
majormoses: Yes, this is the only functional change. The motivation was 100% false positives on any 404 on a clean install of tomcat as in the message (included in screenshot) it will contain the words you are looking for. It has a nice side benefit of performance but that was not the motivation.

Hmm thats an interesting thought about a backdoor returning a 404 but I dont see any way to get around that with the current implementation and would a massive change in order to discover and would probably need some kind of server access to really truly vet its present and they are clever like that.

andresriancho: Ok, now I understand the root cause! Thanks for explaining it to me.

I just created this issue: https://github.com/andresriancho/w3af/issues/15103 to track and fix the problem.

This PR will be closed without merging since, even though it does fix the issue, also potentially reduces the situations in which a backdoor is found.

majormoses:
@andresriancho makes sense to me, should we open a pr to rev the dependency so the next person who tries contributing and uses pgp key signing does not have an issue?

andresriancho:
not sure what you're talking about, could you please explain?

majormoses:
regarding git util. If you commit with a signed key (pgp) it blows up when you try to do things again as the dependencies don't have support for git commit signing.

andresriancho:
https://github.com/andresriancho/w3af/issues/15156

Created that issue to follow up

PR #14800, opened 2016-12-15
Performance and code quality improvement for fuzzy_string_cmp.py

floyd-fuh:
Basically while implementing a Burp plugin that does string comparison similar to the code in w3af/core/controllers/misc/fuzzy_string_cmp.py (which I contributed parts of several years ago) I realized that UPPER_BOUNDS is completely unnecessary, as it can be described by (2.0*len(a))/(len(a)+len(b)) instead. This pull request removes all the bloat of UPPER_BOUNDS, implements the new formula and makes the code a lot more readable.

Please test and let me know what you think :)


andresriancho:
Changes look good

Tests pass

Merging!

PR #15253, opened 2017-03-03
Adds form detection to csrf suitability check

niklasfemerstrand:
In order to exploit a CSRF vulnerability a target URL must contain forms. This change extends the suitability check of audit.csrf to determine if the URL has forms.

### DIFF
--- a/w3af/plugins/audit/csrf.py
+++ b/w3af/plugins/audit/csrf.py
@@ -32,6 +32,8 @@ from w3af.core.controllers.misc.fuzzy_string_cmp import relative_distance_boolea
 from w3af.core.data.fuzzer.fuzzer import create_mutants
 from w3af.core.data.fuzzer.mutants.headers_mutant import HeadersMutant
 from w3af.core.data.kb.vuln import Vuln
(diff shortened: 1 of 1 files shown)


andresriancho:
I believe this doesn't work how you expect, or maybe I'm missing something! Lets take a look at this scenario:

 * Fuzzable request (`freq` in the call to `is_suitable`) is a POST request to `/launch-missile`
 * After sending the request to the action, the HTML says: "Thanks for destroying the enemy"
 * `response` just contains the string mentioned above. No forms exist in this response.
 * The POST request might still be vulnerable to CSRF

If you confirm this, then I believe we should close this PR; but initiate a different conversation to follow-up on this idea. Maybe instead of checking if the HTML in the response has an HTML form we could check if the `hosted_at_url` (https://github.com/andresriancho/w3af/blob/master/w3af/core/data/parsers/utils/form_params.py#L103) contains one.

Still, it wouldn't be a 1:1 match, because w3af could identify a POST request that is vulnerable to CSRF and the HTML has no forms (but the javascript run created dynamic forms, an onclick, etc.). I'm planning on implementing JS crawling support soon and we have to take these scenarios into account.

PR #15285, opened 2017-03-09
[Fix] escape char

hillstonenet-interns-bj:
When writing a back slash in the command line (without any character after), the application used to crash.

Comes from a return after catching an exception (in the parse_Line function). Just added another return for the case where an exception appears.


andresriancho:
Thanks for the fix! I merged it in 390600d1a53963e70d39d9dc1827dc80d2e75451

Had to manually merge the fix because it also included other changes by me (192f2fc)

PR #15392, opened 2017-03-30
Adds a control to avoid auditing CSS and JS files

oguzhantopgul:
Current csrf plugin audits CSS and JS files and results false positives.
I think "_is_suitable" method is the right place to check that. 
Please take a look and let me know what you think


andresriancho:
This was recently implemented in the csrf plugin.

Thanks for the PR and sorry to ignore it for so long.

PR #15429, opened 2017-04-04
Improvements to audit/os-commanding plugin

ilyaglow:
Hi there!

This PR improves detection of the command execution vulnerabilities by adding new line character separator for the commands generation in `audit/os-commanding` plugin. [Commix actually use it](https://github.com/commixproject/commix/blob/d7d0d90e0e7d07da55ba38115d8cd08880d24e0a/src/utils/settings.py#L188).

Also mutants list are extended with candidates in a form: `original parameter value + separator + payload`, so not only `separator + payload` candidates will be used for fuzzing.

Related real-world case (should be definitely more out there):
- [RaXnet Cacti 0.5/0.6.x/0.8.x - Graph_Image.php Remote Command Execution Variant](https://www.exploit-db.com/exploits/25927/)


andresriancho:
I'm very happy with this PR, simple and adds coverage to detect more vulnerabilities.

The only thing that makes me doubt is that the number is HTTP requests sent per parameter to test is multiplied by 2.4 (aprox). The 2x comes from the `append=True` and the 0.4x (aprox) comes from the new lines.

Did you think about this? Ideas? Acceptable by users?

ilyaglow:
Thanks for review @andresriancho!

In my case it's acceptable, but I think your concern is totally right - the plugin should definitely have a couple of options to tune the burden of requests.

Not sure what *level* should be set to default. As it was? Or with just new line characters separators? What settings you think users would be comfortable with?

andresriancho:
What do you think of https://github.com/andresriancho/w3af/issues/15457 ? With that feature we would prevent "all plugins" from having a lot of options to enable and disable which payloads to send.

ilyaglow:
Looks great! Nice way to build consistency across plugins. I'm all for it.

andresriancho:
Manually merged your changes, without the new configuration options for the plugin. This will force me to implement the other changes :+1: 

Want to give https://github.com/andresriancho/w3af/issues/15457 a try? Might be a good chance to learn how to make a complex change in w3af.

ilyaglow:
That would be a good opportunity to dig a bit deeper, I'm in! Just gonna grab a friday beer and see what I can do :)

PR #15711, opened 2017-06-14
Fixes zlib header check decompression exception

niklasfemerstrand:
Some deflate targets which use a different zlib header triggers incorrect header check exceptions:

    Error -3 while decompressing data: incorrect header check

### DIFF
--- a/w3af/core/data/url/handlers/gzip_handler.py
+++ b/w3af/core/data/url/handlers/gzip_handler.py
@@ -57,7 +57,7 @@ class HTTPGzipProcessor(urllib2.BaseHandler):
                     data = gzip.GzipFile(fileobj=StringIO(body)).read()
                 elif "deflate" in enc_hdr:
                     body = response.read()
-                    data = zlib.decompress(body)
+                    data = zlib.decompress(body, -zlib.MAX_WBITS)
                 else:
                     data = response.read()
             except:


w3af/core/data/url/handlers/gzip_handler.py
```diff
@@ -57,7 +57,7 @@ def http_response(self, request, response):
                     data = gzip.GzipFile(fileobj=StringIO(body)).read()
                 elif "deflate" in enc_hdr:
                     body = response.read()
-                    data = zlib.decompress(body)
+                    data = zlib.decompress(body, -zlib.MAX_WBITS)
```
andresriancho: I'm afraid of this line introducing a regression in the gzip handler. What if this fixes the case you're facing but breaks other cases? The zlib window size, as explained in the [python docs](https://docs.python.org/2/library/zlib.html#zlib.decompress) is a little bit difficult to understand and I'm not sure if this change will allow w3af to decompress other bodies.

Maybe the best bet is to a) try to decompress using `data = zlib.decompress(body)` and if that fails with the exception `Error -3 while decompressing data: incorrect header check` (be specific when catching this exception) we try b) `data = zlib.decompress(body, -zlib.MAX_WBITS)`

Do you think this would work?

qnrq: I considered this as well. Initially I tried to set the wbits to 0, which according to the manual automatically determines the window size of the header, but for reasons unknown to me the same exception was thrown.

I have not seen any target that I could potentially test against besides the one that caused the problem. Do you know of any targets which use deflate? It would make it easier to test.

I also read somewhere that gzip could be used as a drop-in replacement of zlib in these cases, however upon further examination I discovered that gzip uses zlib. Perhaps it would still be an issue if handled by gzip.

PR #15712, opened 2017-06-14
Adds RFC1951 uncompression when RFC1950 fails

niklasfemerstrand:
Some deflate targets which use a different zlib header triggers incorrect header check exceptions:

   Error -3 while decompressing data: incorrect header check

Related discussion: https://github.com/andresriancho/w3af/pull/15711#pullrequestreview-43999001


andresriancho:
Thanks for the inspiration to create https://github.com/andresriancho/w3af/commit/2eb5c9237e45b0b6042d1cd12858e829c7f79142

And sorry for not acting on this PR before

PR #15728, opened 2017-06-21
Create JSON file output plugin

paralax:
"Export identified vulnerabilities to a JSON file."

started with same information and schema as the CSV output, extended


w3af/plugins/output/json_file.py
```diff
@@ -0,0 +1,145 @@
...
+                        "VulnDB ID": info.get_vulndb_id(),
+                        "Severity": info.get_severity(),
+                        "Description": info.get_desc()}
+                items.append(item)
+            except Exception, e:
+                msg = ('An exception was raised while trying to write the '
+                       ' vulnerabilities to the output file. Exception: "%s"')
+                om.out.error(msg % e)
+                output_handler.close()
+                print(e)
```
andresriancho: should be removed

w3af/plugins/output/json_file.py
```diff
@@ -0,0 +1,145 @@
...
+                return
+
+        json.dump(items, output_handler)
```
andresriancho: use indent=4 so humans can also read this doc

w3af/plugins/output/json_file.py
```diff
@@ -0,0 +1,145 @@
...
+        self.output_file = os.path.expanduser(self.output_file)
+
+        try:
+            output_handler = file(self.output_file, 'wb')
+        except IOError, ioe:
+            msg = 'Failed to open the output file for writing: "%s"'
+            om.out.error(msg % ioe)
+            return
+
+        items = []
```
andresriancho: I think that at least we should have two items in the json document root:
 * scan info
 * items

The scan info should be similar to the XML, containing the target URL and (also important) the JSON document version. This last bit of information is helpful for parsers to identify how to handle the json document

Items should have all the vulns

andresriancho (commented):
Thanks for the PR! Looks really good, added a couple of things which would improve it a bit and make it ready for merging.

andresriancho:
@paralax this looks awesome! Before merging I'll write a unittest to make sure this works as expected and will continue to work in the future.

PS: If you want to, you can try to write the unittest yourself. Inspiration:
https://github.com/andresriancho/w3af/tree/master/w3af/plugins/tests/output

paralax:
drafted a unit test based on the XML output unit test. have not been able to get it to run cleanly in my local dev env. see if that helps. 

i've been using this plugin to dump results to a lightweight machine readable format for further processing. thanks for w3af and making it easy to extend.

andresriancho:
Wohooooo! We have a test!

I just pushed your changes into the develop branch. Had to modify the test a little bit, but everything is working now :+1: 

You are awesome, thanks!

PS: If you ever want to run the tests on your workstation, just do this:

```
cd /w3af/tests/
docker-compose up
```

In another console:

```
cd /w3af/tests/
sudo ./add-test-routes.sh
```

That should make the `moth` test application (and a bunch of others) available for you.

PR #15894, opened 2017-08-01
Test on a more current Python 2

cclauss:
Do you consider to add [flake8](http://flake8.readthedocs.io) to the testing?  It seems quite confused by the underscores...

flake8 testing of https://github.com/andresriancho/w3af on Python 2.7.13

$ __flake8 . --count --select=E901,E999,F821,F822,F823 --show-source --statistics__
```
./w3af/core/controllers/core_helpers/exception_handler.py:75:22: F821 undefined name '_'
                    (_, _, exception_data.traceback),
                     ^

./w3af/core/controllers/core_helpers/exception_handler.py:75:25: F821 undefined name '_'
                    (_, _, exception_data.traceback),
                        ^

./w3af/core/controllers/exception_handling/helpers.py:103:22: F821 undefined name '_'
    crash_dump.write(_('Submit this bug here:'
                     ^
__...__
```

### DIFF
--- a/circle.yml
+++ b/circle.yml
@@ -9,7 +9,7 @@ machine:
     - docker
 
   python:
-    version: 2.7.3
+    version: 2.7.13
 
   post:
     # This was required to avoid issues with different builds of python being


andresriancho:
I'm a little bit lost here, could you help me out?

I understand that it would be beneficial to run our tests using a newer version of python. That is a change that I would accept, but we need to make sure that things don't break in the process.

On the other hand, does any of w3af's build steps run flake8? I believe no. How does flake8 come into play here?

cclauss:
Sure...  Flake8 is a linter which finds code smells that are not easy to spot with the human eye.  It is the kind of that that could be added to the `test:` part of your Circle CI process.  Repos like [Django do this](https://github.com/django/django/blob/master/tox.ini#L10) to catch errors on all pull requests.  Flake8 can be a lifesaver when doing a refactor, moving to a new library, or moving from Python 2 to Python 3, etc.

The five error conditions (E901,E999,F821,F822,F823) that I had Flake8 focus on are particularly nasty because they throw exceptions that will halt the program if they are not caught.  I am sure that `_()` is defined somewhere but neither I nor flake8 could not figure where that was, thus my question.

andresriancho:
re: flake8, we [use pylint for that](https://github.com/andresriancho/w3af/blob/a53d59b73c493013a90fccf468cf443e47f869d0/w3af/core/controllers/tests/test_pylint.py). Unless there is a very strong reason to migrate to flake8, I would avoid it since:

 * We customized pylint for our project [using pylint plugins](https://github.com/andresriancho/w3af/tree/a53d59b73c493013a90fccf468cf443e47f869d0/w3af/core/controllers/tests/pylint_plugins). I guess we would have to rewrite those for flake8
 * Our code is annotated with `# pylint-ignore ...` to remove false positives detected by pylint. I suspect that flake8 will have another set of false positives and won't support pylint's annotations.

re: `_`, yes, I solved that here in [pylintrc](https://github.com/andresriancho/w3af/blob/a53d59b73c493013a90fccf468cf443e47f869d0/w3af/core/controllers/tests/pylint.rc):

```
# List of additional names supposed to be defined in builtins. Remember that
# you should avoid to define new builtins when possible.
additional-builtins=_
```

The underscore comes from the gettext module https://docs.python.org/2/library/gettext.html#gettext.gettext

cclauss:
Perfect.  Thanks for the explanation.  Should I close this PR?

PR #15909, opened 2017-08-07
UnicodeDecodeError running find_dvcs .git

Gxsghsn:
i only test git_index, so i did't change others, may test tomorrow


andresriancho:
Thanks for this! Quick question, are contents of that file always encoded in utf-8?

Gxsghsn:
and because of w3af/plugins/output/console.py _make_printable string.printable , w3af can not print unicode to stdout...

andresriancho:
This change is not longer needed in the latest version of the plugin, which does this in a more generic way for all the filenames generated by all DVCS extractors

https://github.com/andresriancho/w3af/blob/5550a5f35070a3c28989ad1472c83ec98a7e5f91/w3af/plugins/crawl/find_dvcs.py#L109-L135

PR #15987, opened 2017-08-31
meta content generator grep plugin

paralax:
sometimes i just want to peek at the meta tag that specifies a site's generator and version, often used by CMS systems. this plugin does just that, which wasn't available in the meta tags plugin (nor did it seem to fit).

includes a test case.


w3af/plugins/tests/grep/test_meta_generator.py
```diff
@@ -0,0 +1,122 @@
...
+from w3af.core.data.parsers.doc.url import URL
+from w3af.core.data.dc.headers import Headers
+from w3af.core.controllers.misc.temp_dir import create_temp_dir
+from w3af.plugins.tests.helper import PluginTest, PluginConfig
+from w3af.plugins.grep.meta_generator import meta_generator
+
+
+class TestMetaGenerator(PluginTest):
+
+    meta_generator_url = get_moth_http('/grep/meta_generator/')
```
andresriancho: Did you make any modifications to moth? Will this test PASS without them?

w3af/plugins/tests/grep/test_meta_generator.py
```diff
@@ -0,0 +1,122 @@
...
+        self.assertEquals(set(['Interesting META tag'] * 2),
+                          set([v.get_name() for v in vulns]))
+
+        joined_desc = ''.join([v.get_desc() for v in vulns])
+
+        self.assertIn('linux', joined_desc)
+        self.assertIn('verify-v1', joined_desc)
+
+
+class TestMetaGeneratorRaw(unittest.TestCase):
```
andresriancho: Awesome, this is more than enough of a test for me.

w3af/plugins/tests/grep/test_meta_generator.py
paralax: i did not, no. as such this test might _fail_. hmm ...

andresriancho:
Thanks for the inspiration and initial code for the new plugin commited at https://github.com/andresriancho/w3af/commit/66f782f5edb7ef1f5ee1f72ec752b05a4b23314c

Kept you as one of the plugin authors

Issue #14966, opened 2017-01-15
Fix dependencies on newer systems


derboblan:
##### Steps to reproduce the problem:

1. Run Debian testing (probably other distros affecteda aswell)
2. clone repo
3. running w3af_dependency_install.sh breaks pip and other stuff, w3af can't be started

##### Solution
Bump Version of pyOpenSSL in w3af/core/controllers/dependency_check/requirements.py to 16.2.0

now you can run

virtualenv --clear --system-site-packages env
. bin/env/activate
pip install -I ... (whatever the dependency check tells you to)

without proplems

##### System information

PR #16037, opened 2017-09-29
Bumping pyOpenSSL version

dav009:
Connects to #14966
credits to @jrlonan


andresriancho:
Just updated this yesterday, thanks!

PR #16063, opened 2017-10-07
Fixing race condition when terminating grep consumer

linerd0196:
When user stop the scan by calling `w3afCore.stop()` method, a race condition problem of grep consumer can cause the scan never stops. 
`w3afCore.stop()` call `consumer.terminate()` method, which consume all remaining items on it's `in_queue`. It check the queue's emptiness using `empty()` method. 

However, this method is marked 'not reliable' by Queue module's documentation.  In this case, the race condition happens between the check and the calling of`in_queue.join()` method. Then the `in_queue.join()` method never stops because the queue is not really empty at that moment and will never be consumed again.

My solution is that we create a new thread, which consumes `in_queue` until the consumer is marked `stopped`, while in the current thread, the queue is joined, and returns immediately when it's empty. When the `join()` method return, we mark the consumer `stopped` to finish the thread we created before.

Please check it out and tell me what you think.

### DIFF
--- a/w3af/core/controllers/core_helpers/consumers/base_consumer.py
+++ b/w3af/core/controllers/core_helpers/consumers/base_consumer.py
@@ -79,12 +79,13 @@ class BaseConsumer(Process):
         """
         :param consumer_plugins: Instances of base_consumer plugins in a list
         :param w3af_core: The w3af core that we'll use for status reporting
-        :param thread_name: How to name the current thread
+        :param thread_name: How to name the current thread, eg. Auditor
         :param create_pool: True to create a worker pool for this consumer
         """
         super(BaseConsumer, self).__init__(name='%sController' % thread_name)
 
-        self.in_queue = QueueSpeed(maxsize=max_in_queue_size)
+        self.in_queue = QueueSpeed(maxsize=max_in_queue_size,
+                                   name=thread_name)
         self._out_queue = Queue.Queue()
         
         self._consumer_plugins = consumer_plugins
--- a/w3af/core/controllers/core_helpers/consumers/base_consumer.py
+++ b/w3af/core/controllers/core_helpers/consumers/base_consumer.py
@@ -1,17 +1,23 @@
 """
 base_consumer.py
+
 Copyright 2012 Andres Riancho
+
 This file is part of w3af, http://w3af.org/ .
+
 w3af is free software; you can redistribute it and/or modify
 it under the terms of the GNU General Public License as published by
 the Free Software Foundation version 2 of the License.
+
 w3af is distributed in the hope that it will be useful,
 but WITHOUT ANY WARRANTY; without even the implied warranty of
 MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
 GNU General Public License for more details.
+
 You should have received a copy of the GNU General Public License
 along with w3af; if not, write to the Free Software
 Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
 """
 import Queue
 import sys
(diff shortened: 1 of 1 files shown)


andresriancho:
> When user stop the scan by calling w3afCore.stop() method, a race condition problem of grep consumer can cause the scan never stops.
w3afCore.stop() call consumer.terminate() method, which consume all remaining items on it's in_queue. It check the queue's emptiness using empty() method.

I believe you're talking about this code from `base_consumer.py`?

```python
    def terminate(self):
        """
        Remove all queued work from in_queue and poison the loop so the consumer
        exits. Should be very fast and called only if we don't care about the
        queued work anymore (ie. user clicked stop in the UI).
        """
        while not self.in_queue.empty():
            self.in_queue.get()
            self.in_queue.task_done()
        
        self.join()
```

What is going on here is:

 * The while loop empties the queue. Any work that was queued is removed and marked as done.
 * Yes, there might be a race condition after the while and before calling `self.join()`. The race condition would happen if any other thread adds a new item to the queue

Now, lets go to `self.join()` implementation:

```python
    def join(self):
        """
        Poison the loop and wait for all queued work to finish this might take
        some time to process.
        """
        if not self.is_alive():
            # This return has a long history, follow it here:
            # https://github.com/andresriancho/w3af/issues/1172
            return

        if not self._poison_pill_sent:
            # https://github.com/andresriancho/w3af/issues/9587
            self._poison_pill_sent = True
            self.in_queue_put(POISON_PILL)

        self.in_queue.join()
```

Keep in mind that while all this is running, the `BaseConsumer.run` method is also running in a different thread.

Now to the `join` code, first we make sure the thread `is_alive` (`run` is running), then we make sure that the poison pill is sent only once (maybe a race condition there too if the `join` method is called twice by different threads) and finally we join the `in_queue`.

So, when we enter `join` with one or two tasks in the queue, the `join` method adds the `POISON_PILL` and then waits for the `run` method to do its job: consume the tasks, consume the poison pill and die.

The issue with calling `empty()` in `terminate` doesn't seem to be so bad.

andresriancho:
How would your changes improve this situation?

linerd0196:
> Yes, there might be a race condition after the while and before calling self.join(). The race condition would happen if any other thread adds a new item to the queue

No. The race condition does not happens there. It happens between `C1` and `C2` in the following code, which runned by `grep` consumer. 

```python
    def join(self):
        """
        Poison the loop and wait for all queued work to finish this might take
        some time to process.
        """
        if not self.is_alive():
            # This return has a long history, follow it here:
            # https://github.com/andresriancho/w3af/issues/1172
            return

        if not self._poison_pill_sent:
            # https://github.com/andresriancho/w3af/issues/9587
            self._poison_pill_sent = True
            self.in_queue_put(POISON_PILL)            # C1

        self.in_queue.join()                          # C2
```

After `C1`, if any fuzzable request is put into grep's queue, C2 will never end. Because grep's `run()` method will stop right after it get POISON_PILL. Following items will not be processed. So the queue will never be empty and the `join()` call will never stop.

The chance a fuzzable request is put into grep's queue after `C1` is significant. If I'm not wrong, whenever a HTTP request is sent, It and it's response will be sent to grep consumer as a FuzzableRequest object.

My changes will guarantee that all item on grep's queue will be processed, therefore the `join()` call will always be `joinable`

Another solution I think is that we check `self._poison_pill_sent` on `in_queue_put()` method. So other threads couldn't send an item after POISON_PILL is sent.

```python
    def in_queue_put(self, work):
        if self._poison_pill_send and work is not None:
            return self.in_queue.put(work)
```
 
Should I create a new pull request?

linerd0196:
My claim about `empty()` in `terminate` is wrong :((

andresriancho:
> After C1, if any fuzzable request is put into grep's queue, C2 will never end. Because grep's run() method will stop right after it get POISON_PILL, and items after POISON_PILL will not be processed. So the queue will never be empty, but C2 waits for the queue to be empty. Therefore the join() call will never stop.

Ahh! Nice catch!

> The chance a fuzzable request is put into grep's queue after C1 is significant

Completely agree

andresriancho:
> Another solution is that we check self._poison_pill_sent on in_queue_put() method. So other threads couldn't send anything to the queue after POISON_PILL is sent. I think it's more efficient than the first solution.

I like this solution much more, easier to read and understand.

Please do modify this PR so that the new solution is implemented instead. Also, I would prefer something like this:

```python
    def in_queue_put(self, work):
        # A long comment explaining why this is here, reference to this PR, etc.
        if self._poison_pill_sent:
            return

        if work is not None:
            return self.in_queue.put(work)
```

linerd0196:
PR changed 👍

andresriancho:
Amazing, thanks for your hard debugging and understanding of the code. Sometimes people don't understand how much work is related with a two line change: :clap: :clap: :clap: :clap:

PR #16091, opened 2017-10-22
Clean REST api's start scan codes

linerd0196:
When user starts scan from REST api, 2 w3afCore instances will be created, one by `start_scan()` function in `w3af/core/ui/api/resources/scans.py`, another by `start_scan_helper()` function in `w3af/core/ui/api/utils/scans.py`.
I think creating 2 cores is not necessary. Both cores are initialized (threadpools are created), but when the scan stops, only the second w3afCore on `w3af/core/ui/api/utils/scans.py` is cleaned up, the first isn't. That will lead to a situation when threads created by the first core will never stop

### DIFF
--- a/w3af/core/ui/api/resources/scans.py
+++ b/w3af/core/ui/api/resources/scans.py
@@ -26,7 +26,8 @@ from flask import jsonify, request
 from w3af.core.ui.api import app
 from w3af.core.ui.api.utils.error import abort
 from w3af.core.ui.api.utils.auth import requires_auth
-from w3af.core.ui.api.db.master import SCANS
+from w3af.core.ui.api.db.master import SCANS, ScanInfo
+from w3af.core.ui.api.utils.log_handler import RESTAPIOutput
 from w3af.core.ui.api.utils.scans import (get_scan_info_from_id,
                                           start_scan_helper,
                                           get_new_scan_id,
@@ -34,6 +35,7 @@ from w3af.core.ui.api.utils.scans import (get_scan_info_from_id,
 from w3af.core.data.parsers.doc.url import URL
 from w3af.core.controllers.w3afCore import w3afCore
 from w3af.core.controllers.exceptions import BaseFrameworkException
+import w3af.core.controllers.output_manager as om
 
 
 @app.route('/scans/', methods=['POST'])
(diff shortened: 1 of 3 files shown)


andresriancho:
Oh, this is something that shouldn't happen. I'll look into it later today when I have more time.

w3af/core/ui/api/resources/scans.py
```diff
@@ -102,21 +104,28 @@ def start_scan():
...
+    # Add the REST API output plugin
+    om.manager.set_output_plugin_inst(scan_info.output)
+
     #
     # Finally, start the scan in a different thread
     #
-    scan_id = get_new_scan_id()
-    scan_info_setup = Event()
-
-    args = (target_urls, scan_profile, scan_info_setup)
```
andresriancho: Why did you remove this Event? I don't remember exactly why I added it, but it seems that at some point it made sense to me to use it to:

```python
# Wait until the thread starts
scan_info_setup.wait()
```

andresriancho:
I reviewed and agree with most of the PR, added a question regarding the `Event`.

Will merge after I understand that.

linerd0196:
I think the reason why you add that `Event` is to guarantee the `scan_info` object is added to SCANS before returning HTTP response to user. The code which creates `scan_info` runs in different thread (`start_scan_helper`), and the `Event` object is lock in `start_scan` thread until the `scan_info` object is added successfully in `start_scan_helper` thread.

``` python
def start_scan_helper(target_urls, scan_profile, scan_info_setup):
    """
    Create a new instance of w3afCore, save it to SCANS and run core.start()
    :param scan_profile: The contents of a profile configuration
    :param scan_info_setup: Event to set when the scan started
    :return: The instance of w3afCore.
    """
    scan_info = ScanInfo()
    SCANS[get_new_scan_id()] = scan_info
    scan_info.w3af_core = w3af_core = w3afCore()
    scan_info.target_urls = target_urls
    scan_info.output = RESTAPIOutput()

    scan_info_setup.set()

    scan_profile_file_name, profile_path = create_temp_profile(scan_profile)
```

In this PR, there is no reason to keep the `Event` object, because the `scan_info` is added to `SCANS` before we create a new thread which runs the real start.

andresriancho:
Awesome, thanks for explaining this to me :+1:

PR #16184, opened 2017-11-26
enhanced global_redirect plugin

Gxsghsn:
enhanced global_redirect plugin

### DIFF
--- a/w3af/plugins/audit/global_redirect.py
+++ b/w3af/plugins/audit/global_redirect.py
@@ -36,8 +36,8 @@ class global_redirect(AuditPlugin):
     Find scripts that redirect the browser to any site.
     :author: Andres Riancho (andres.riancho@gmail.com)
     """
-    TEST_URLS = ('http://www.w3af.org/',
-                 '//w3af.org')
+    TEST_URLS = ['http://www.w3af.org/',
+                 '//w3af.org']
 
     def __init__(self):
         AuditPlugin.__init__(self)
@@ -54,6 +54,17 @@ class global_redirect(AuditPlugin):
 
         :param freq: A FuzzableRequest object
         """
+
+        netloc = freq._uri.netloc
+        netloc = netloc.split(':')[0]
+
+        self.TEST_URLS.extend([netloc + '.w3af.org',
+                               '//' + netloc + '.w3af.org',
+                               'http://' + netloc + 'w3af.org/',
+                               netloc + '@w3af.org',
+                               '//' + netloc + '@w3af.org',
+                               'http://' + netloc + '@w3af.org/'])
+
         mutants = create_mutants(freq, self.TEST_URLS)
         
         self._send_mutants_in_threads(self._uri_opener.send_mutant,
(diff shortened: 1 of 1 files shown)


w3af/plugins/audit/global_redirect.py
```diff
@@ -54,6 +55,17 @@ def audit(self, freq, orig_response):
 
         :param freq: A FuzzableRequest object
         """
+
+        netloc = freq._uri.netloc
```
andresriancho: I would use `freq.get_uri()` instead of `freq._uri`, you never know what the getter might do

w3af/plugins/audit/global_redirect.py
```diff
@@ -54,6 +55,17 @@ def audit(self, freq, orig_response):
...
+        netloc = netloc.split(':')[0]
+
+        self.TEST_URLS.extend([netloc + '.' + self.TEST_DOMAIN,
+                               '//' + netloc + '.' + self.TEST_DOMAIN,
+                               'http://' + netloc + '.' + self.TEST_DOMAIN + '/',
```
andresriancho: If we add this one, shouldn't we also add one with `https://`? Or maybe chose dynamically based on the protocol used by the fuzzable request?

w3af/plugins/audit/global_redirect.py
```diff
@@ -54,6 +55,17 @@ def audit(self, freq, orig_response):
...
+                               netloc + '@'+ self.TEST_DOMAIN,
+                               '//' + netloc + '@' + self.TEST_DOMAIN,
+                               'http://' + netloc + '@'+ self.TEST_DOMAIN + '/'])
```
andresriancho: Same as above

w3af/plugins/audit/global_redirect.py
```diff
@@ -112,8 +124,7 @@ def _domains_are_equal(self, redir_url, test_url):
         """
         try:
             redir_domain = urlparse.urlparse(redir_url).netloc
-            test_domain = urlparse.urlparse(test_url).netloc
-            return redir_domain == test_domain
+            return redir_domain.endswith(self.TEST_DOMAIN)
```
andresriancho: I don't like this, what if the site redirects to `http://target.com.w3af.org/`? Note the trailing slash

andresriancho:
I like this PR, it adds a lot of value to the plugin and is well thought. Please fix the comments I've made and I'll merge it right away.

w3af/plugins/audit/global_redirect.py
Gxsghsn: `redir_domain` is processed by `urlparse.urlparse()`, so there should not a trailing slash

w3af/plugins/audit/global_redirect.py
Gxsghsn: this i did not use dynamically way, because some site only sso is `https`, so it will redirect from `http` to `https` or the other way round, just like `https://sso.xxx.com/login?service=http%3A%2F%2Fwww.test.com%2F`

andresriancho: Then, shouldn't we have both, http and https?

Not thinking only about the SSO scenario, but all web apps.

w3af/plugins/audit/global_redirect.py
andresriancho: Ahh you are right, sorry about that.

andresriancho:
Awesome, thanks!

PR #16219, opened 2017-11-29
fix xss json header with charset

Gxsghsn:
some site response json header is `Content-Type: application/json;charset=UTF-8`, so i split the content_type to fit


w3af/plugins/audit/xss.py
```diff
@@ -158,6 +158,9 @@ def _identify_trivial_xss(self, mutant):
         ct_options, _ = response.get_headers().iget('X-Content-Type-Options')
         content_type, _ = response.get_headers().iget('Content-Type')
 
+        if content_type is not None:
+            content_type = content_type.split(';')[0].strip()
```
andresriancho: I would add a comment above this line to explain why it was added.

Also, did you try to search through the rest of the plugins for `Content-Type` to see if similar issues exist?

w3af/plugins/audit/xss.py
```diff
@@ -107,8 +112,9 @@ def _check_xss_in_parameter(self, mutant, debugging_id):
         """
         Tries to identify (persistent) XSS in one parameter.
         """
-        if not self._identify_trivial_xss(mutant):
-            self._search_xss(mutant, debugging_id)
+
+        if self._identify_trivial_xss(mutant) == False:
```
andresriancho: Why break something that was ok? :-)

https://www.python.org/dev/peps/pep-0008/ says:

> Don't compare boolean values to True or False using ==.
> Yes:   if greeting:
> No:    if greeting == True:
> Worse: if greeting is True:

w3af/plugins/audit/xss.py
```diff
@@ -150,19 +156,15 @@ def _identify_trivial_xss(self, mutant):
         
         response = self._uri_opener.send_mutant(trivial_mutant)
 
+        # special character caused a json response,
+        # return None to close the later search
+        if self._is_json_response(response):
+            return None
```
andresriancho: If the code gets to this point, and `self._check_persistent_xss` is True, then we're not running:

```
self._xss_mutants.append((trivial_mutant, response.id))
```

I wouldn't change the order of these calls, just in case.

Gxsghsn:
thanks for your review. i noticed you didn't check json response in _analyze_echo_result, and this cause some misinformation 
`A Cross Site Scripting vulnerability was found at: "http://test.com/xss/xss.php", using HTTP method GET. The sent data was: "a=xsfgi%3Cxsfgi" The modified parameter was "a".This vulnerability was found in the request with id 41.`

andresriancho:
Thanks for your ideas and code!

A part of your code was already merged into `xss.py` a long time ago, but I just made some modifications to merge all of your code.

PR #16317, opened 2017-12-20
add support for WPScan plugins usage

paralax:
the WPScan project does a fantastic job of maintaining a list of plugins and their vulnerabilities. but i hate using two tools when one can work. so i wrote a plugin to use their JSON file.
- if needed, it downloads the plugins.json file from their github and stores it locally
- it parses it
- it scans a target site for wordpress plugins
- if found, it records a piece of info or, if it has known vulns, a vuln entry

### DIFF
--- a/w3af/plugins/crawl/wpscan.py
+++ b/w3af/plugins/crawl/wpscan.py
@@ -21,8 +21,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 """
 import json
 import os
-import urllib
-import zipfile
 
 from itertools import repeat, izip
 
@@ -64,7 +62,6 @@ class wpscan(CrawlPlugin):
         self._exec = True
         self._already_tested = DiskSet(table_prefix='wpscan')
 
-    @runonce(exc_class=RunOnce)
     def crawl(self, fuzzable_request):
         """
         Get the file and parse it.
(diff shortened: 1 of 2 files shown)


w3af/plugins/crawl/wpscan.py
```diff
@@ -0,0 +1,232 @@
...
+                om.out.debug("done")
+            with open('/tmp/data.zip', 'wb') as f:
+                f.write(urllib.urlopen('https://github.com/wpscanteam/wpscan/raw/master/data.zip').read())
+            _wpscan_data_zip = zipfile.ZipFile(open('/tmp/data.zip', 'rb'))
+            _wpscan_data_zip.extract('data/plugins.json', os.path.join(self.BASE_PATH))
+            os.rename(os.path.join(self.BASE_PATH, 'data', 'plugins.json'), 
+                      os.path.join(self.BASE_PATH, 'plugins.json'))
+            om.out.debug('done')
+        self._plugin_list = json.load(open(os.path.join(self.BASE_PATH, 'plugins.json'), 'r'))
+        if not self._exec:
```
andresriancho: How does this interact with the `@runonce` decorator?

w3af/plugins/crawl/wpscan.py
```diff
@@ -0,0 +1,232 @@
...
+        if self._update_plugins or not os.path.exists(os.path.join(self.BASE_PATH, 'plugins.json')):
+            om.out.debug("Downloading the archive from wpscan and extracting plugins.json ...", )
+            try:
+                os.stat(BASE_PATH)
+            except OSError:
+                om.out.debug("%s doesn't exist, making ..." % BASE_PATH)
+                os.mkdir(BASE_PATH)
+                om.out.debug("done")
+            with open('/tmp/data.zip', 'wb') as f:
+                f.write(urllib.urlopen('https://github.com/wpscanteam/wpscan/raw/master/data.zip').read())
```
andresriancho: Not so important: This wouldn't work in an environment with a proxy config, even if the user configured the proxy in w3af.

Rather important: This is not the usual way in which we do things in w3af. I prefer including the `plugins.json` file in our repository and have everything in one place ready for the user to download and use. What is the license for `plugins.json`?

There are other plugins, like phishtank for example, which have a unittest which checks the date of the data file and fails if the date is too old. This makes sure that we update the data file for the unittest to PASS -> for the build to PASS.

andresriancho:
Awesome work, thanks!

ethicalhack3r:
Hi guys! Just a quick note on the licensing of the WPScan data.

The data license is the same as the software, which can be found here https://raw.githubusercontent.com/wpscanteam/wpscan/master/LICENSE

Please include the license file if you plan on including the data files. Our free license prohibits commercial use, which may make it incompatible with w3af's license, I'm not sure.

Any questions let me know!

andresriancho:
@paralax thanks for your changes!

@ethicalhack3r I totally respect the license you used and as w3af's project lead I do not intend to go against you / wpscan team licensing decisions. Also let me add that I personally would have chosen a different license, but that is a different discussion we don't really need to have.

fact: w3af uses GPL2.0

fact: wpscan license is incompatible with GPL2.0 (as far as I can understand it, wpscan license restricts / goes against some of the liberties that GPL2.0 grants).

Please confirm, but it seems that including wpscan's JSON file in w3af would:
 * a) break w3af's licensing: it wouldn't be a "pure" GPL2.0 software
 * b) violate wpscan licensing

IANAL but if any of these is true then we can't include the file.

Also, not including the file in the repository, but downloading the file on each use would potentially prevent us from a) but most likely not from b).

Finally, it seems that w3af is already violating the wpscan license by using this file:

https://github.com/andresriancho/w3af/blob/master/w3af/plugins/crawl/wordpress_fingerprint/wp_versions.xml

Thoughts?

paralax:
I have no interest in either violating the wpscan license and harming that teams hard work, or in ruining the w3af project. Let's not merge this then. I'll find another way to build a Wordpress plugin scanner for w3af. 

Thank you both.

andresriancho:
@ethicalhack3r what are your thoughts on this?

paralax:
closing due to license incompatibilities.

PR #16331, opened 2017-12-28
Add docker creds

kalzoo:
Add the docker credentials (which are given on the "advanced" page) for quicker setup.


andresriancho:
Hmmmm, not sure which version of the docker image you are using, but for me running 

```
sudo ./extras/docker/scripts/w3af_console_docker
```

Doesn't require me to enter a password. If it is required I agree that the password should be in the docs!

How can I reproduce this issue?

Issue #16371, opened 2018-01-05
Unexpected EOF with invalid request methods on HTTPS targets


qnrq:
```OpenSSL.SSL.SysCallError: (-1, 'Unexpected EOF')```

```
The allowed_methods plugin got an error while requesting "https://X". Reason: "Unexpected EOF"
The file_upload plugin got an error while requesting "https://X". Reason: "Unexpected EOF"
The server_header plugin got an error while requesting "https://X". Reason: "Unexpected EOF"
```

caused when plugins make requests with invalid methods, e.g. ```TRACK```, ```ARGENTINA```, ```INDEX```.

andresriancho:
Thanks for reporting this issue!

Before we dive into the fix, what is the root cause for it? Which servers are affected? Can we reproduce this with `curl`?

I suspect that this is caused when a specific server receives an HTTP request with an unknown method and the connection is abruptly closed by the server side?

Maybe an approach to fix this issue is to mark all requests generated by the `dav` plugin (the one sending those invalid methods) with `error_handling` and handling any exceptions in the plugin?

andresriancho:
@qnrq ping!

PR #16372, opened 2018-01-05
Handles OpenSSL errors when making deliberate invalid requests

niklasfemerstrand:
https://github.com/andresriancho/w3af/issues/16371


andresriancho:
This one was fixed in a different way

PR #16520, opened 2018-01-31
Added '-y' and '--yes' flags to allow the user to skip the disclaimer.

chadduffin:
Allowing the user to automatically acknowledge and skip agreeing to the disclaimer allows the W3AF tool to be more easily used with scripts.

This PR is a single commit that adds two flags, '-y' and '--yes', that facilitate this automatic acknowledgement.

### DIFF
--- a/w3af_console
+++ b/w3af_console
@@ -51,6 +51,9 @@ Options:
     -v or --version
         Show w3af's version
 
+    -y or --yes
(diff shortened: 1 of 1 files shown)


andresriancho:
Thanks!

chadduffin:
Thank you!

PR #16621, opened 2018-02-14
Adds max_discovery_urls misc setting

niklasfemerstrand:


andresriancho:
Thanks for the PR! A few things about it:

 * It seems that the changes from strategy.py are missing?

 * In this area of the code the `_report_max_time` attribute is used to make sure that the message is only reported once. Maybe the same is needed for the `max_discover_urls` feature? 

https://github.com/qnrq/w3af/blob/75dff0060c9b4f531b403f2eb3c0790260a06e6f/w3af/core/controllers/core_helpers/consumers/crawl_infrastructure.py#L327-L333

 * In these lines, the user won't know what the number being printed next to the plugin name is:

https://github.com/qnrq/w3af/blob/75dff0060c9b4f531b403f2eb3c0790260a06e6f/w3af/core/controllers/core_helpers/consumers/crawl_infrastructure.py#L424-L429

I believe that we shouldn't print this data as information(), it should be printed as debug(). A debug line like:

```
Identified 123 out of 500 URLs defined by max_discover_urls.
```

Would have more meaning for users. Also, this line should only be printed when `max_discover_urls` is set.

PR #17143, opened 2018-07-19
audit ssl_certificate: support Server Name Indication (SNI)

mattjbray:
This adds a new parameter `ssl_certificate.sniHostname`.


andresriancho:
Thanks for the PR, having SNI support for this is important.

A few questions and comments on the PR:
 * Why did you use a user-configurable parameter when you had the `domain` variable? https://github.com/mattjbray/w3af/blob/a5c01e6d21749da624330456e829d3ac2cc79902/w3af/plugins/audit/ssl_certificate.py#L121 This would have worked really well and doesn't require the user to set the user-configured parameter:
`python
ssl_sock = ssl_context.wrap_socket(s, server_hostname=domain)
`
 * What happens if you try to scan a site that does NOT support SNI with this code?

andresriancho:
Latest w3af implements this in a better way (see ssl_wrapper.py)

PR #17166, opened 2018-08-05
Vulners Scanner plugin release

vulnersCom:
This plugin greps every software banner and checks vulnerabilities online at vulners.com database.
By default it's using **anonymous** Vulners API entry point. 
If you will get 418 errors (rate limit fired) you can obtain API key for free at https://vulners.com.
Then configure it at vulners_scanner as vulners_api_key variable.

It's using [Vulners Python API](https://github.com/vulnersCom/api)


w3af/plugins/grep/vulners_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+        self._vulnerability_cache = {}
+
+    def get_options(self):
+        """
+        :return: A list of option objects for this plugin.
+        """
+        ol = OptionList()
+
+        d = ('Vulners API key for extended scanning rate limits.'
+             'You can obtain one for free at https://vulners.com')
```
andresriancho: Grammar / Style: ` Obtain a free API key for free at https://vulners.com/`

Also, note the leading space in the string, that will add a space after the dot from the previous line.

andresriancho: What happens if the user enters an invalid API key? Where will the plugin crash?

w3af/plugins/grep/vulners_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def set_options(self, options_list):
+        """
+        This method sets all the options that are configured using the user
+        interface generated by the framework using the result of get_options().
+        :param options_list: A dictionary with the options for the plugin.
+        :return: No value is returned.
+        """
+        self._vulners_api_key = options_list['vulners_api_key'].get_value()
+
```
andresriancho: Remove one new line after this method.

w3af/plugins/grep/vulners_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+    def grep(self, request, response):
+        """
+        Plugin entry point, search for vulnerable software banners  in web application HTTP responses.
+        :param request: The HTTP request object.
+        :param response: The HTTP response object
+        :return: None
+
+        """
+        # Lazy update rules if it's first start of the plugin
+        if not self.rules_updated:
```
andresriancho: There is a race condition here, where multiple threads will be running the the rules will be updated 2 or 3 times.

Use (check in other plugins, not 100% sure about the name):

```python
with self._plugin_lock:
    if not self.rules_updated:
        ...
```

w3af/plugins/grep/vulners_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+        """
+        self._vulners_api_key = options_list['vulners_api_key'].get_value()
+
+
+    def update_vulners_rules(self):
+        # Get fresh rules from Vulners.
+        try:
+            self.rules_table = self.get_vulners_api().rules()
+            # Adapt it for MultiRe structure [(regex,alias)] removing regex duplicated
+            regexAliases = collections.defaultdict(list)
```
andresriancho: Use _ instead of camel case.

regex_aliases
software_name

w3af/plugins/grep/vulners_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+            return
+        # We do not parse non-text output
+        if not response.is_text_or_html():
+            return
+        if response.get_url().get_domain_path() in self._already_visited:
+            return
+
+        self._already_visited.add(response.get_url().get_domain_path())
+
+        raw_response = response.dump()
```
andresriancho: Do you guys need to match the whole HTTP response? Including the body? What if we just match the regex against headers?

My fear is that this will at some point leak sensitive information to vulners.com. How?
 * A broken regex is obtained by w3af in `update_vulners_rules`. 
 * This regex matches more than required and matches sensitive information (cookie, response body, etc.)
 * w3af plugin sends that to vulners.com in `check_vulners` `software_version` parameter

I would feel much better if we could just send the HTTP response headers to `self._multi_re.query` below, and before doing that we clean the sensitive headers like set-cookie.

w3af/plugins/grep/vulners_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+
+            v = Vuln(name= bulletin['id'],
+                     desc = bulletin['description'] or bulletin.get('sourceData', bulletin['title']),
+                     severity = self.get_severity(bulletin.get('cvss', {}).get('score', 0)),
+                     response_ids = response.id,
+                     plugin_name = self.get_name())
+
+            v.set_url(response.get_url())
+
+            self.kb_append_uniq(location_a = self,
```
andresriancho: Let's say that the HTTP response header shows a `Server` which triggers a vulnerability detection. An old vulnerable web server is in use in this application. All HTTP responses generated by the application include this server header and thus trigger the vulnerability detection. Will the scan end with 293810 findings (one for each URL) for this vulnerability?

I haven't run a scan to confirm, but it looks like that.

I believe you guys need to research https://github.com/andresriancho/w3af/blob/master/w3af/core/data/kb/info_set.py and group by vulnerability. In other words, report an `InfoSet` vulnerability with title: `Old Vulnerable Server` and then add `Info` instances to that `InfoSet`

w3af/plugins/grep/vulners_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+        return self._vulners_api
+
+    def check_vulners(self, software_name, software_version, check_type):
+        vulnerabilities = {}
+        cached_result = self._vulnerability_cache.get((software_name, software_version, check_type))
+        if cached_result:
+            return cached_result
+        # Ask Vulners about vulnerabilities
+        if check_type == 'software':
+            vulnerabilities = self.get_vulners_api().softwareVulnerabilities(software_name, software_version)
```
andresriancho: What happens when the API triggers a 418 here?

Same two lines below in call to `cpeVulnerabilities`

Shouldn't we log at least the error so the user knows that he needs an API key?

andresriancho: Also, when 418 is triggered for the first time, shouldn't we disable the grep plugin? If we hit the limit then it makes no sense to keep trying to match any regexes.

w3af/plugins/grep/vulners_scanner.py
```diff
@@ -0,0 +1,219 @@
...
+                                info = v,
+                                filter_by = 'URL')
+
+
+    def get_long_desc(self):
+        """
+        :return: A DETAILED description of the plugin functions and features.
+        """
+        return """
+        This plugin greps every software banner and checks vulnerabilities online at vulners.com database.
```
andresriancho: > This plugin greps software banners and checks vulnerabilities online at vulners.com database.

> By default the grep plugin uses anonymous Vulners API entry point (rate-limited to XXX requests per minute / hour / day), it is possible to get a free API key at https://vulners.com/ to avoid rate limits.

> Configure the API key using the vulners_api_key user-configured parameter.

andresriancho (commented):
Many small comments. Mostly concerned about 418 and information leaks.

w3af/plugins/grep/vulners_scanner.py
vulnersCom: Grammar fixed.
Nope, it will just report verbose error "Invalid key..e.t.c" by the Vulners lib.

w3af/plugins/grep/vulners_scanner.py
vulnersCom: Got it.

w3af/plugins/grep/vulners_scanner.py
vulnersCom: Wow! Fantastic scenario, never thought about it.
We are using this rules: https://github.com/vulnersCom/detect-rules
They also matches http body and JS scripts and so on. Not only headers.
I think the best way will be:
1) You will fork them to w3af github account.
2) Plugin will take them from yours github.
3) We will make time-to-time update of them, and you will pull updates validating them.

I've understand that it will take a little time to maintain, but users will bu sure what kind of regexes are used. Privacy first, yep.

w3af/plugins/grep/vulners_scanner.py
vulnersCom: API can't hit 418 when using Vulners Python API.
There is throttling inside. It reads response headers from the Vulners telling client API to slow down.

w3af/plugins/grep/vulners_scanner.py
vulnersCom: Fixed.

vulnersCom: 10rps at the anonymous mode.
(b.t.w it's more than enough for the most users)

w3af/plugins/grep/vulners_scanner.py
vulnersCom: Applied.

w3af/plugins/grep/vulners_scanner.py
vulnersCom: Fixed.

w3af/plugins/grep/vulners_scanner.py
vulnersCom: Got it. Fixed.

vulnersCom:
I've updated the plugin following your comments.
Please, take a look.

The only unfixed part is:
Regex collection download location. This one is up on you.
If you ok I will add new loc for it.

w3af/plugins/grep/vulners_scanner.py
andresriancho: Where is that verbose error printed? If stdout, then only users running w3af_console and looking at the console will notice that. The error will never be saved to a log file, XML file, etc. which is bad.

The library should raise an exception, which the plugin should catch and log using `om.out.error`.

The error should be printed to the log only once and then the plugin should be disabled.

w3af/plugins/grep/vulners_scanner.py
andresriancho: Sadly, and because of licensing issues, I can't include rules.json file in the w3af repository.

rules.json is GPLv3
w3af is GPLv2

Those licenses are incompatible. I can't distribute GPLv3 licensed files inside my GPLv2 project.

A potential solution would be for w3af to download https://raw.githubusercontent.com/vulnersCom/detect-rules/master/rules.json at the beginning of the scan. You could do it similar to https://github.com/andresriancho/w3af/blob/develop/w3af/plugins/grep/retirejs.py#L148-L167

I trust https://raw.githubusercontent.com/vulnersCom/detect-rules/master/rules.json more than the result of the vulners.com API because:
 * There is a git log which we could potentially use to find someone modifying the JSON
 * Users can audit it if they want to, knowing that whatever is in that URL can not be changed unless the git log changes

vulnersCom: Ok! Will do it that way.
But looks like I need to checkout at develop branch, because it master grep plugin does not have "self._uri_opener".

vulnersCom:
I will close this pull req. 
Without tuning GrepPlugin itself it will not work.
I will create pull req to the develop branch.

PR #17193, opened 2018-08-16
Update the crawl.open_api plugin with additional parameters

artem-smotrakov:
First of all, thanks for adding `crawl.open_api` which allows scanning REST APIs. Currently the plugin discovers API endpoints by parsing Open API specification which should be available on a target. If API spec was found in one of the predefined location, `crawl.open_api` plugin then validates the specification.

Unfortunately it sometimes happens that an application doesn't provide API specification. Furthermore, sometimes applications provide API specs which may be not quite valid, and as a result, validation fails. These two problems with applications may prevent w3af to scan them. Ideally, applications should provide a valid specification in one of the well-known locations but sometimes updating applications may take time and block scans. To prevent such delays, and make `crawl.open_api` a bit more flexible, I would like to propose the following parameters:

- `no_spec_validation` for disabling Open API validation
- `custom_spec_location` for providing a local path to Open API spec

I am having some troubles with running tests, so no tests so far.


artem-smotrakov:
Sorry for creating a pull request against master. Closing it.

PR #17235, opened 2018-08-31
fix bug with detached terminal

lekster:
When run w3af with detached terminal (by cron for example), a lot of errors was generated
the commit fix this problem

`./w3af_console -s ./scripts/sqli.w3af < /dev/null

Unknown key. The following commands are allowed during the scan:

(P) pause the scan
(R) resume a paused scan
(enter) print scan status
(Ctrl+C) stop scan

Unknown key. The following commands are allowed during the scan:

(P) pause the scan
(R) resume a paused scan
(enter) print scan status
(Ctrl+C) stop scan

Unknown key. The following commands are allowed during the scan:

(P) pause the scan
(R) resume a paused scan

`


andresriancho:
Thanks!

PR #17237, opened 2018-08-31
Automatic HTTP header discovery in Open API plugin

artem-smotrakov:
Open API standard allows defining parameters via HTTP headers. Currently w3af doesn't try to test HTTP headers if a user didn't explicitly list them in `misc-settings.fuzzable_headers` configuration parameter. As a result, if a user wants to test HTTP API which receive parameters via HTTP headers, the user has to manually extract header names from the Open API spec and put them to the `misc-settings.fuzzable_headers` configuration parameter.

In this patch, I would like to propose an automated discovery of all parameters which are passed via HTTP headers, and enabling them for further testing:

- Enabled this feature by default if Open API plugin is enabled
- Added new `open_api.discover_fuzzable_headers` configuration parameter to disable the feature

Implementation note: the Open API plugin passes discovered headers to `HeadersMutant` class via a new field in the global config. I am not sure it's the best way, but that's what I could only come up with since I am not an expert in w3af. Please let me know if there is any problem with it.


andresriancho:
My first reaction to this PR is :heart: , it provides great value and it should have been there since the beginning.

I dislike the modifications to `fuzzer/fuzzer.py` where `fuzzable_openapi_headers` (something that has plugin scope) is added to a list of things which are global / framework scope. I understand that there is no good way for doing this without modifying *something* in the framework, but maybe we can find a more elegant solution!

Also, I believe that all w3af users will want to use this feature, so... maybe... don't add the flag to disable it? 

I completely :heart: that you added unittests!

To solve the problem at hand: Headers are not fuzzed unless the user adds them to the config. What do you think about modifying this piece of code https://github.com/andresriancho/w3af/blob/master/w3af/core/data/fuzzer/mutants/headers_mutant.py#L49-L66

Maybe we can do something like this:
 * The `FuzzableRequest` class has a new pair of methods: `set_force_fuzzing_headers(headers_list)` and `get_force_fuzzing_headers()`
 * The OpenAPI parser will call `set_force_fuzzing_headers(headers_list)`  with the headers it finds
 * The `HeadersMutant.create_mutants()` gets the headers to apply fuzzing to, not only from the config, but also from `get_force_fuzzing_headers()`

When doing this, keep in mind that `FuzzableRequest` is a very core part of w3af, and you'll have to pay attention to:
 * `FuzzableRequest.__slots__`
 * `FuzzableRequest.from_parts`

Let me know if this idea makes sense!

artem-smotrakov:
Thank you for your feedback @andresriancho !

There are two reasons why I added the new configuration parameter for the Open API plugin:

- A user may want to tests only a limited set of headers, so that the user can disable the feature, and specify a list of headers in `misc-settings.fuzzable_headers`. Maybe there is another way to do that.
- It may be good to have a convenient way to disable the feature just in case. For example, if one finds in unnecessary for a particular applications which we didn't consider, or maybe the user discovers a bug, etc.

I don't like the solution with global `fuzzable_openapi_headers` either. Let me look into your proposal. At first glance, it looks better then the original solution. I also need to get a better understanding how `FuzzableRequest` works. Thanks!

w3af/core/data/fuzzer/fuzzer.py
```diff
@@ -104,6 +104,7 @@ def create_mutants(freq, mutant_str_list, append=False, fuzzable_param_list=[],
 
 
 CONF_KEYS = [('fuzzable_headers', []),
+             ('fuzzable_openapi_headers', []),
```
artem-smotrakov: I'll remove it.

artem-smotrakov:
I implemented a draft of the solution with `FuzzableRequest.set_force_fuzzing_headers(headers_list)`. I am planning to do some testing, add some docs, and do some cleanup. Here are a couple of comments:

1. I added `_force_fuzzing_headers` to `__slots__` but I didn't really measure how it improves memory usage and performance.
2. If I am not missing something, not sure that  `_force_fuzzing_headers` should be handled in `FuzzableRequest.from_parts`
3. `FuzzableRequest` doesn't let others to modify `_force_fuzzing_headers` - not a bit deal probably, but may save us from some bugs.

@andresriancho Could you please take a look if we're heading the right direction?

w3af/core/data/request/fuzzable_request.py
```diff
@@ -429,6 +434,29 @@ def set_headers(self, headers):
...
+        """
+        if headers is None:
+            raise TypeError('headers should not be null')
+
+        if not isinstance(headers, collections.Iterable):
+            raise TypeError(TYPE_ERROR % ('_force_fuzzing_headers', 'iterable'))
+
+        self._force_fuzzing_headers = set()
+        for header in headers:
+            self._force_fuzzing_headers.add(header)
```
andresriancho: Why like this and not: `self._force_fuzzing_headers = set(headers)`?

If you do it like this, it is not possible for someone to remove a header from the set once it has been set.

andresriancho:
Just added a small comment. If you run some live tests and tell me it works, I'll merge.

w3af/core/data/request/fuzzable_request.py
artem-smotrakov: `self._force_fuzzing_headers = set(headers)` looks better - I just overcomplicated this :) I'll update it. Thanks!

artem-smotrakov:
@andresriancho Please take a look at the latest diff. I did some live testing, and it seems to work fine. Please let me know if you have questions.

artem-smotrakov:
@andresriancho Could you please take a look at it again when you have some time? I see changes coming to the plugin, if you don't have objections, we can merge and avoid possible merging issues. Thanks!

andresriancho:
I was looking into this one more time and everything looks good except from one thing (please correct me if I'm mistaken with the way I read the code).

If the openapi specification has two endpoints /a and /b , and only /a is using headers (per the specification), when requests to /b will also include the header that is defined for /a. Correct?

Also, like you see for other parts of the code where we parse the endpoint parameters, it would be nice to parse the default value (if it exists) and set it in the FuzzableRequest. If this is not done, w3af will most likely send HTTP requests similar to:

```
GET /a?some-parameter={payloads-here} HTTP/1.1
Custom-Header: {empty}
```

While trying to fuzz the `some-parameter` parameter. This is bad because there might be a vulnerability in `some-parameter` which we're not reaching because a validation in `Custom-Header` that could be "bypassed" if we send the default value.

Just for you to understand the other side of the story, when fuzzing the `Custom-Header` the requests will look like:

```
GET /a?some-parameter={default} HTTP/1.1
Custom-Header: {payloads-here}
```

Where `{default}` sent in `some-parameter` give us more chances of finding the vulnerabilities that are in `Custom-Header`.

artem-smotrakov:
> If the openapi specification has two endpoints /a and /b , and only /a is using headers (per the specification), when requests to /b will also include the header that is defined for /a. Correct?

I think requests to /b should not include headers which are specific for /a. Let me check it.

> Also, like you see for other parts of the code where we parse the endpoint parameters, it would be nice to parse the default value (if it exists) and set it in the FuzzableRequest.

According to https://swagger.io/docs/specification/2-0/describing-parameters/ , the default value is the one that the server uses if the client does not supply the parameter value in the request. If I understand correctly, if a parameter is not required then there are two ways a fuzzer can go:
- just skip it (don't send an empty value), so that the target should use the default value
- send the default value (if defined), and if it's not defined then skip this header (don't send an empty value)

Let me check how w3af behaves, and fix it if necessary. Both ways look okay for me. What do you think?

> While trying to fuzz the some-parameter parameter. This is bad because there might be a vulnerability in some-parameter which we're not reaching because a validation in Custom-Header that could be "bypassed" if we send the default value.

It totally makes sense to me. I am wondering if we can extract even more from spec, so that we could enumerate more cases. For example, I am wondering if we could use values defined in enums. It may be a next improvement for API scans.

andresriancho:
> According to https://swagger.io/docs/specification/2-0/describing-parameters/ , the default value is the one that the server uses if the client does not supply the parameter value in the request. 

Yeah... hmmm... maybe we should force it. Some (80%) of the apps might follow that spec, but the rest will do anything else they want. So, if there is a default in the OpenAPI spec then the HTTP requests going out to that endpoint should include it. We don't want to trust the app behaving like the spec.

My fear here is that:
 * The openapi spec states a default
 * w3af says: "it is ok not to send this parameter because the default will be used"
 * w3af doesn't send the param
 * The application does some basic validation on the request and says: "the parameter X is not there, stop"
 * There was a vulnerability in parameter Y which was sent and is being actively tested using payloads.

So we better always send X with the default parameter we have for it.

artem-smotrakov:
Good point. It makes sense to me. Let me try to update it to use default values if present. Right now I am trying to add a test for it, but I see that something is failing on calling `copy.deepcopy()`

https://github.com/andresriancho/w3af/blob/356b14b975039706f4fd7f4f5db5b114cd75f14e/w3af/core/data/parsers/doc/open_api/parameters.py#L62

artem-smotrakov:
I have updated the patch:

- Now requests to /b do not include headers which are specific for /a
- If a header has a default value, it's always set in the FuzzableRequest. Please see `test_json_multiple_paths_and_headers()` test case which shows how it works now.
- Updated tests
- Fixed `test_no_model_json_object_with_int_param_in_body()` test case (the test failed before, it's not related to the patch)
- Added `test_disabling_spec_validation()` test case (I owe it from the previous pull request)

A couple of notes:

- If a header parameter doesn't have a default value, then the plugin generates two fuzzable requests with forced headers: one with no valued for the header, and another one with a generated value. See `X-Bar-Header` in `test_disabling_headers_discovery()` test case.
- `_should_skip_setting_param_value()` checks only header parameters for default values, but maybe it can consider other types of parameters.
- Currently `SpecificationHandler._set_operation_params()` considers only two cases. First, it sets values only for required fields. Second, it sets values for both required and optional fields. In both cases, it also sets values for headers if the spec defines default values. We can implement a couple of improvements. For example, we can add a third case when it doesn't force setting default values according to the spec. In this case, we'll let the target to do it. Another possible improvement is to enumerate all values in enums.
- At first, I tried to add a couple of test cases to `w3af/plugins/crawl/open_api.py` but looks like the test is quite unstable.

@andresriancho Could you please take a look? Maybe the notes above can be further improvements. I can try to find time to work on them.

andresriancho:
Awesome! I'll try to take a look at this today

andresriancho:
Just read the code, run the tests, everything looks perfect.

Thank you so much for the great feature and interactions (good communication is an integral part of contributing to open source, and not all of us can do it well).

giphy

artem-smotrakov:
Thank you for the review and feedback :)

PR #17241, opened 2018-09-04
fix type problem in clean_data_container

lekster:
fix error - AttributeError: 'int' object has no attribute 'isdigit' when value is int
I will can fix it in most wide case like this, but i'm not sure:

        if isinstance(value, str):
            if value.isdigit():
                _type = 'number'
            else:
                _type = 'string'
        else:
            _type = type(value)


lekster:
sorry, but i not understand :)
I catched the problem when i tried to fuzzing by JsonContainer where token_filter method always return true and i have int fields in my json

What about this if str(value).isdigit()  ?

PR #17243, opened 2018-09-05
add skip-dependencies-check param to startup.conf

lekster:
In some reason i need run w3af without dependencies check in console mode, i think the best way is to set new param in startup.conf

### DIFF
--- a/w3af/core/controllers/dependency_check/dependency_check.py
+++ b/w3af/core/controllers/dependency_check/dependency_check.py
@@ -171,7 +171,7 @@ def dependency_check(dependency_set=CORE, exit_on_failure=True):
     :return: True if the process should exit
     """
     if StartUpConfig().get_skip_dependencies_check():
-        return True
+        return False
 
     disable_warnings()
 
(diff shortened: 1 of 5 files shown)


andresriancho:
I like the code, and the refactoring you made to make it clearer / easier to read.

The only problem I see is in `w3af_console`, where the call to `dependency_check()` was moved from line 13 to somewhere around line 100. The `dependency_check()` should be before [all these imports](https://github.com/andresriancho/w3af/blob/master/w3af_console#L15-L17)

Moving the call to any place after those imports will break the feature introduced by the function: dependency check tries to give the user a friendly message when a dependency is not imported. If you move it after other w3af imports, what will happen is that those imports will fail (if the user doesn't have all dependencies installed) and show an ugly traceback.

andresriancho:
Tried to find a good way to make this work without making the code ugly (parsing the `startup.conf` before `dependency_check()` would be an option... but looks ugly...) and was unable to find a way to do it.

lekster:
how about of hiding the check functionality in dependency_check.py ? (see update above)

andresriancho:
Seems to be working fine, thanks!

andresriancho:
giphy

PR #17301, opened 2018-10-02
clean_data_container() should expect None value

artem-smotrakov:
I noticed that an NPE may occur in `clean_data_container()` function because `value` may be None:

`AttributeError: 'NoneType' object has no attribute 'isdigit'`

If I understand correctly, it's a valid case when `value` is None. For example, it may happen if a JSON payload contains a null. How about adding a check if `value` is None? I was also wondering if it should check that `key` is not null?


andresriancho:
Awesome, thanks!

PR #17307, opened 2018-10-04
Don't use deepcopy() in ParameterHandler to make debugging great again

artem-smotrakov:
While running a scan against a RESTful application, I noticed that debugging doesn't work well sometimes. In particular, the debugger couldn't stop on a breakpoint under some circumstances. Then, I found the following error messages in logs which seem to cause the problem:

> Exception RuntimeError: 'maximum recursion depth exceeded in __subclasscheck__' in <type 'exceptions.RuntimeError'> ignored
Exception RuntimeError: 'maximum recursion depth exceeded in __subclasscheck__' in <type 'exceptions.RuntimeError'> ignored

It turned out that the errors above are caused by calling `deepcopy()` on Bravado's objects in `ParameterHandler`. It looks similar to https://stackoverflow.com/questions/32831050/pycharms-debugger-gives-up-when-hitting-copy-deepcopy

Updating bravado-core to 5.0.7 didn't help.

In this patch, I'd like to propose getting rid of `deepcopy` in `ParameterHandler`. It tried to replace it with `Operation.from_spec()`. It seems to work well, and fixes the problem with debugging.

https://github.com/Yelp/bravado-core/blob/master/bravado_core/operation.py#L121

### DIFF
--- a/w3af/core/data/parsers/doc/open_api/parameters.py
+++ b/w3af/core/data/parsers/doc/open_api/parameters.py
@@ -20,9 +20,10 @@ along with w3af; if not, write to the Free Software
 Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 """
-import copy
 import random
 import datetime
+
+from bravado_core.operation import Operation
 
 from w3af.core.data.fuzzer.form_filler import (smart_fill,
                                                smart_fill_file)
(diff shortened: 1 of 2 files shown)


andresriancho:
Amazing, thanks!

I was completely unaware of the debugging problems associated with deepcopy

PR #17311, opened 2018-10-08
add target url param for w3af_console

lekster:
I needed to run w3af_console with a same script on a set of target urls,
this commit adds the "target" parameter

### DIFF
--- a/w3af_console
+++ b/w3af_console
@@ -54,6 +54,9 @@ Options:
     -y or --yes
         Automatically agree to disclaimer prompt
 
+    -t or --target
+        Target URL for scan
+
 For more info visit http://w3af.org/
 """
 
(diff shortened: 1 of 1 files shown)


w3af_console
```diff
@@ -75,16 +78,19 @@ def _usage():
...
     user interface at startup.
     
     :param script_file: The script (-s) file name
     :param profile: The profile (-p) name
     :param force_profile: Should we run the profile? (-P)
+    :target_url: Target URL for scan
     """
     commands_to_run = []
+    if target_url:
+        commands_to_run.extend(['target', 'set target {}'.format(target_url), 'back'])
```
andresriancho: Please validate `target_url` using a white-list to make sure it only contains `0-9a-zA-z:/@-.?&=` (and maybe some other chars that are valid in URLs and I'm forgetting).

Do this to prevent "w3af console command injection" when someone specifies a `target_url` containing new lines / carriage returns.

andresriancho:
Thanks!

giphy

PR #17323, opened 2018-10-13
Update scapy depedency

gpotter2:
Hello !

Scapy-real is a super outdated project, that was only updated once in 2013
The real version is `scapy`, and it now contains all features that were temporarily held in scapy-real.

We are planning to break the scapy-real package on PyPi to avoid confusion


andresriancho:
Thanks for sending the PR and letting us know!

gpotter2:
FTR

`pip install scapy-real` will now fail with a deprecation warning

Issue #17315, opened 2018-10-11
Enhance the OpenAPI plugin to accept user-provided values for parameters


artem-smotrakov:
While generating fuzzable requests for API endpoints, the OpenAPI plugin tries to find good values for the parameters which are accepted by the API endpoints. The plugin considers default values provided in the OpenAPI spec, or tries to guess the best value based on parameter name, type, format, etc. I think this is a very good approach for an initial scan but it may not be enough if we want to test an application a bit deeper. The problem is that there is not an easy way to let the plugin know about values it should use for testing. Let's consider the following API endpoint:

```
paths:
  /users/{userId}:
    post:
      summary: Update user
      parameters:
        - in: path
          name: userId
          schema:
            type: integer
          required: true
          description: Numeric ID of the user to get
        - in: header
          name: X-User-Name
          schema:
            type: string
          required: true
          description: User name
```

In this case, the OpenAPI plugin creates a fuzzable request which assigns some pre-defined number (for example, 42) to `userId` and some pre-defined string to `X-User-Name`. The plugin tries its best when guessing valid values, but it has no idea about the actual context. When w3af tries to fuzz `X-User-Name`, it will send requests to `/users/42`. Most probably the requests will be rejected if no used exists with this ID. As a result, w3af won't actually test how the application processes the `X-User-Name` header.

I would like to propose adding a way how a user can provide a list of valid values for specific parameters for API endpoints. Currently I don't see the plugin provides such a way. The feature would help to use w3af in a better and more efficient way for:

- grey-box testing when a tester already has some knowledge about the application
- regular testing applications in CI/CD

There are only a couple of workarounds which are not flexible:

- use custom API spec which defines single default values
- use `headers_file -> header_file` option to set pre-defined headers

I think it would be much better if a user could provide a list of meaningful values for a particular parameter. Basically, a user needs to specify a mapping `{path, parameter_name} -> [ list of values ]`

I can implement such a feature if it gets a positive feedback. Before opening a pull request, I'd like to agree on a way how a user can provide such values. I would add `parameter_values` option for the OpenAPI plugin. The option should contain a path to a file which contains `{path, parameter_name} -> [ list of values ]` mapping. The file may be in JSON, YAML or CSV format. I'd start with JSON or YAML.

Please let me know what you think.

andresriancho:
[...]

PR #17339, opened 2018-10-16
Enhance the OpenAPI plugin to accept user-provided values for parameters

artem-smotrakov:
This is a patch for #17315 

Main changes:

- Added `parameter_values_location` configuration option for the OpenAPI plugin. The option allows to set a path to a YAML file which contains context-specific parameter values.
- Added `ParameterValues` class which can load context-specific parameter values from a YAML file. The file contains a mapping `{ path, parameter name } -> list of values` . Currently, it doesn't use an HTTP method but it can be easily updated. For `in: body` parameter type, a user can specify a JSON payload, but it's not currently possible to specify a value for a particular field in the JSON payload.
- Updated `ParameterHandler` to use context-specific parameter values. If a user provided custom parameter values, then the class prefers them, and tries to enumerate all possible combinations of the specified parameters.
- Fixed a possible null-dereference in `get_uri()` method.
- Added a couple of tests.

I am testing the feature on a couple of applications but would like to start a review now to get feedback earlier.


artem-smotrakov:
I have tested it with a couple of applications - it worked well.

w3af/core/data/parsers/doc/open_api/parameters.py
```diff
@@ -33,6 +35,83 @@ class OpenAPIParamResolutionException(Exception):
...
+        """
+        with open(filename, 'r') as content:
+            return self.load_from_string(content)
+
+    def load_from_string(self, string):
+        """
+        Loads parameter values from YAML.
+        :param string: Definition of parameter values in YAML.
+        """
+        content = yaml.load(string)
```
andresriancho: User might enter an invalid YAML, that case should be handled.

w3af/plugins/crawl/open_api.py
```diff
@@ -407,6 +415,13 @@ def get_options(self):
         o = opt_factory('discover_fuzzable_headers', self._discover_fuzzable_headers, d, BOOL, help=h)
         ol.add(o)
 
+        d = 'Path to a file with parameter values.'
+        h = ('This option sets a path to a YAML file which contains parameter values'
+             ' which should be used in testing API endpoints. If no parameter values are provided,'
+             ' the plugin tries to guess them.')
+        o = opt_factory('parameter_values_location', self._parameter_values_location, d, INPUT_FILE, help=h)
```
andresriancho: Maybe the YAML format validation should be done here? That could be done by changing the `INPUT_FILE` type to `YAML_INPUT_FILE` and creating the logic for handling that.

It should be fairly easy, take a look at:
 * https://github.com/andresriancho/w3af/blob/master/w3af/core/data/options/input_file_option.py
 * https://github.com/andresriancho/w3af/blob/356b14b975039706f4fd7f4f5db5b114cd75f14e/w3af/core/data/options/tests/test_input_file_option.py
 * https://github.com/andresriancho/w3af/blob/356b14b975039706f4fd7f4f5db5b114cd75f14e/w3af/core/data/options/opt_factory.py#L43

andresriancho: `parameter_values_location` should be changed to `parameter_values_file` (that is how input files are usually named)

https://github.com/andresriancho/w3af/blob/356b14b975039706f4fd7f4f5db5b114cd75f14e/w3af/plugins/output/text_file.py#L297
https://github.com/andresriancho/w3af/blob/356b14b975039706f4fd7f4f5db5b114cd75f14e/w3af/plugins/output/xml_file.py#L153

w3af/plugins/crawl/open_api.py
```diff
@@ -448,10 +464,35 @@ def get_long_desc(self):
...
+                - John
+                - Bill
+            - path: /users
+              parameters:
+              - name: user-id
+                values:
+                - 1234567
+              - name: birth-date
+                values:
+                - 2000-01-02
```
andresriancho: I believe that this feature requires its own documentation in https://github.com/andresriancho/w3af/blob/develop/doc/sphinx/scan-rest-apis.rst

I would make the example in `get_long_desc` shorter, and reference users to the documentation with something like: "For more information and examples about this feature read the framework documentation at https://docs.w3af.org"

andresriancho (commented):
Awesome!

andresriancho:
I need to work harder on merging your PR. If I don't do that I'll end up in a branch merge nightmare.

Sorry for not being more responsive with these PRs. I was on vacations. Will try to merge all these amazing things you've been working on to `develop` as soon as the comments are solved.

artem-smotrakov:
Thanks for your comments. Let's work on them one by one, and I'll resolve possible conflicts which may appear. Then, it should be easier to merge them to `develop`.

w3af/core/data/parsers/doc/open_api/parameters.py
artem-smotrakov: Okay, I'll add a try-except block to catch `YAMLError` and wrap the exception in to `ParameterValueParsingError`.

w3af/plugins/crawl/open_api.py
artem-smotrakov: Hmm, that may be nice. Although `input_file_option.py` doesn't look that simple at first glance :) Okay, I'll see what I can do here.

w3af/plugins/crawl/open_api.py
artem-smotrakov: Okay, I'll update both description and docs. In fact, the docs need to be updated with descriptions of the recent changes/features. I hope I can find time to do that.

w3af/plugins/crawl/open_api.py
andresriancho: Yeah, `input_file_option.py` doesn't look simple, but `yaml_file_option.py` should be simple :-)

Take a look at the other files in the same directory, those will give you ideas on how simple the inputs can actually be.

andresriancho: Yes please

artem-smotrakov:
I have updated the patch:

- Renamed configuration parameters in the `crawl.open_api` plugin.
- Added `YAML_INPUT_FILE` option type.
- Updated docs and description for the `crawl.open_api` plugin.
- The `crawl.open_api` plugin now expects an invalid YAML file.
- Added `TestOpenAPICustomParameterValues` for the new configuration option.
- Added `TestOpenAPICustomSpec` for `custom_spec_file` configuration option.

artem-smotrakov:
@andresriancho When you have  some time, could you please take a look at this pull request? I have also updated other pull requests and added several comments - looking for a review. Thanks!

PR #17354, opened 2018-10-22
Add escape chars to brackets in regexes

yvonneKim:
Bug fix for issue #16483 

Added escape chars before left brackets to keep them from being group matched.


andresriancho:
giphy

PR #17355, opened 2018-10-23
Don't force fuzzing auth headers for API endpoints

artem-smotrakov:
I noticed that the OpenAPI plugin forces fuzzing for auth headers. For example, it can happen if the API spec defines authentication with API keys passed in a header. Usually, a user provides such an auth header which then should be included to all HTTP request without any modification. So, it doesn't make much sense to force fuzzing for this header.

This patch fixes this issue. It updates `RequestFactory` to filter out auth headers.


artem-smotrakov:
I missed one thing - let me re-create the pull request.

artem-smotrakov:
Please review https://github.com/andresriancho/w3af/pull/17358

PR #17357, opened 2018-10-23
Update OpenAPI plugin to fuzz URL parts by default

artem-smotrakov:
This is a patch for #17342

Changes:

- Updated the OpenAPI plugin to always turn on `fuzz_url_filenames` and `fuzz_url_parts` options.
- A couple of minor/cosmetic enhancements in `open_api.py`.
- Fixed `TestOpenAPIFindAllEndpointsWithAuth`.
- Added `TestOpenAPIFuzzURLParts`.

PR #17358, opened 2018-10-23
Don't force fuzzing auth headers for API endpoints

artem-smotrakov:
I noticed that the OpenAPI plugin forces fuzzing for auth headers. For example, it can happen if the API spec defines authentication with API keys passed in a header. Usually, a user provides such an auth header which then should be included to all HTTP request without any modification. So, it doesn't make much sense to force fuzzing for this header.

This patch fixes this issue. It updates `RequestFactory` to filter out auth headers.


andresriancho:
Good point, I agree that this shouldn't be fuzzed **for each endpoint**, but shouldn't w3af at least fuzz it for **one** endpoint?

That way, if the REST API has a "SQL injection in the API key verification method" w3af would be able to find it.

artem-smotrakov:
I have updated the patch:

- `RequestFactory` doesn't set headers for fuzzing. This feature is now in `OpenAPI` which I think is a better place for it. 
- Authentication headers which are defined in `securityDefinitions` are forced to be fuzzed only once. It doesn't apply to `Authorization` header - not sure if it should.
- Moved `test_no_forcing_fuzzing_auth_headers` test to `test_main.py`
- Added `TestOpenAPIFuzzAuthHeaders`
- Fixed a bug in `TestOpenAPIFindAllEndpointsWithAuth`. I also fixed it in #17357 but I am going to revert it there. Sorry for possible confusions. I create multiple pull requests to make the changes easier to review. But there are also some test bugs which create some noise. I am trying to get rid of them if possible but the fixes sometimes affect multiple pull requests :)

w3af/core/data/parsers/doc/open_api/main.py
```diff
@@ -178,9 +180,65 @@ def parse(self):
...
+        :param operation: An instance of Operation class
+                          which represents the API endpoint.
+        :return: A list of unique header names.
+        """
+        parameter_headers = set()
+        for parameter_name, parameter in operation.params.iteritems():
+            if parameter.location == 'header':
+
+                # Fuzz auth headers only once.
+                if self._is_already_fuzzed_auth_header(parameter.name, operation.swagger_spec):
```
andresriancho: The parser shouldn't have any responsibility associated with fuzzing. The parser should just return what is found in the OpenAPI, then other parts of the code should be responsible of fuzzing (or not) the authentication header.

Not sure where to put this... maybe the solution is to modify the [fuzzer.py](https://github.com/andresriancho/w3af/blob/master/w3af/core/data/fuzzer/fuzzer.py#L35-L60) code?

Issue #17343, opened 2018-10-17
Create a plugin for checking if all API endpoints require authentication


artem-smotrakov:
It may be good to write a simple plugin which checks that all API endpoints require authentication (I believe in most cases they should be). The plugin can basically send requests to the endpoints without `Authorization` header and expect 401 error. I noticed such issues while [reviewing security configs of Spring-based apps](https://blog.gypsyengineer.com/en/security/tips-configuring-security-rest-api-spring.html). Not sure if it should be implemented in the OpenAPI plugin since it looks like an audit plugin.

artem-smotrakov:
[Here is a couple of comments ](https://github.com/andresriancho/w3af/issues/17315#issuecomment-430260076) from @andresriancho :

> This one looks like a very nice feature, and I agree it should be an audit plugin. Only one comment about this issue:
> 
> - The check should only work when the user configures authentication (api key, headers, etc.)
> - The check should only work when the open api specification requires authentication for an endpoint
> - The check should send a request for each authenticated endpoint using the provided authentication, then send it without the provided authentication and compare both. If they are similar (see fuzzy_equal) then a vulnerability should be reported.

artem-smotrakov:
I'll try to find some time to prototype it.

artem-smotrakov:
I got some time to work on it, and have a prototype which I believe should be ready for review soon. One comment about the following though:

> The check should send a request for each authenticated endpoint using the provided authentication, then send it without the provided authentication and compare both. If they are similar (see fuzzy_equal) then a vulnerability should be reported.

I may be missing something but I am not sure if it's strictly necessary. I think it would be okay (at least for the first version of the plugin) to analyze response codes. If a endpoint requires authentication according to the spec, then it should return 401 if no auth info is present in the request. The plugin can have more complex logic for checking response codes but I am not sure if it should use `fuzzy_equal`. 

I would propose to have a simple logic in the first version, and then we can enhance the plugin, for example:

- Add an option to set expected response codes.
- Add an option to enable comparing responses with `fuzzy_equal`.

andresriancho:
> I may be missing something but I am not sure if it's strictly necessary. I think it would be okay (at least for the first version of the plugin) to analyze response codes. If a endpoint requires authentication according to the spec, then it should return 401 if no auth info is present in the request. The plugin can have more complex logic for checking response codes but I am not sure if it should use fuzzy_equal.

Agreed. Good call.

andresriancho:
[...]

PR #17363, opened 2018-10-26
Add audit.open_api_auth plugin

artem-smotrakov:
This pull request introduces a new plugin `audit.open_api_auth` which is described in #17343 

Implementation notes:

- The new plugin works only with API endpoints. It requires an API specification which has to be provided by `crawl.open_api` plugin.
- The `crawl.open_api` plugin shares API specification via the knowledge base.
- The new plugin simply checks that endpoints return 401 error if no authentication info is provided in the request.

Please consider it as a first version of the plugin. I was going to make the plugin smaller but it turned out that it needs quite a lot of checks, so now the plugin is not that small :) 

I believe the plugin may be improved. Please take a look at several enhancements in the plugin description, and let me know what you think. I am planning to work on them but would prefer to implement them separately after the first version of the plugin is integrated.

### DIFF
--- a/w3af/plugins/audit/open_api_auth.py
+++ b/w3af/plugins/audit/open_api_auth.py
@@ -1,0 +1,113 @@
+"""
+open_api_auth.py
+
+Copyright 2018 Andres Riancho
+
+This file is part of w3af, http://w3af.org/ .
+
+w3af is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation version 2 of the License.
+
+w3af is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with w3af; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+"""
+from w3af.core.controllers.plugins.audit_plugin import AuditPlugin
+
+import w3af.core.data.kb.knowledge_base as kb
+import w3af.core.controllers.output_manager as om
+from w3af.core.data.parsers.doc.open_api.specification import SpecificationHandler
+
+
+class open_api_auth(AuditPlugin):
+    """
+    TODO
+
+    :author: Artem Smotrakov (artem.smotrakov@gmail.com)
+    :author: Andres Riancho (andres.riancho@gmail.com)
+    """
+
+    def __init__(self):
+        AuditPlugin.__init__(self)
+        self._spec = None
+
+    def audit(self, freq, orig_response, debugging_id):
+        """
+        TODO
+
+        :param freq: A FuzzableRequest
+        :param orig_response: The HTTP response associated with the fuzzable request
+        :param debugging_id: A unique identifier for this call to audit()
+        """
+        if self._has_no_api_spec():
+            om.out.info("could not find an API specification, skip")
+            return
+
+        if self._no_security_in_spec():
+            om.out.info("the API spec defines no security, skip")
+            return
+
+
+        # TODO The check should only work when the open api specification requires authentication for a endpoint
+        #      Check for security headers according to the spec
+        if self._should_skip(freq):
+            return
+
+        # TODO The check should send a request for each authenticated endpoint using the provided authentication,
+        # then send it without the provided authentication and compare both.If they are similar (see fuzzy_equal)
+        # then a vulnerability should be reported.
+
+    def _analyze_result(self, mutant, response):
+        """
+        Analyze results of the _send_mutant method.
+        """
+        pass
+
+    # TODO The check should only work when the user configures authentication (api key, headers, etc.)
+    def _no_security_in_spec(self):
+        """
+        TODO
+        :return:
+        """
+        return True
+
+    def _has_no_api_spec(self):
+        """
+        TODO
+        :return:
+        """
+        if self._spec:
+            return False
+
+        specification_handler = kb.kb.raw_read('open_api', 'specification_handler')
+        if not specification_handler:
+            return True
+
+        self._spec = specification_handler.parse()
+
+        if not self._spec:
+            return True
+
+        return False
+
+    def _should_skip(self, freq):
+        """
+        TODO
+        :return:
+        """
+        return True
+
+    def get_long_desc(self):
+        """
+        :return: A DETAILED description of the plugin functions and features.
+        """
+        return """
+        TODO
+        """
--- a/w3af/plugins/audit/open_api_auth.py
+++ b/w3af/plugins/audit/open_api_auth.py
@@ -42,9 +42,11 @@ class open_api_auth(AuditPlugin):
         and keep the rest of the request untouched (headers, payload, etc).
         But on the other hand, the application is supposed to check HTTP method
         in the very beginning, and reject requests if the method is not supported.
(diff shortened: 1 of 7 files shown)


w3af/plugins/audit/open_api_auth.py
```diff
@@ -0,0 +1,372 @@
...
+        returned by the server.
+
+    :author: Artem Smotrakov (artem.smotrakov@gmail.com)
+    :author: Andres Riancho (andres.riancho@gmail.com)
+    """
+
+    def __init__(self):
+        AuditPlugin.__init__(self)
+        self._spec = None
+        self._expected_codes = [401]
```
andresriancho: Maybe add 403?

w3af/plugins/audit/open_api_auth.py
```diff
@@ -0,0 +1,372 @@
...
+        Check if an API endpoint requires authentication according to its API spec.
+
+        :param freq: A FuzzableRequest to the API endpoint.
+        :param orig_response: The HTTP response associated with the fuzzable request.
+        :param debugging_id: A unique identifier for this call to audit().
+        """
+        # The check works only with API endpoints.
+        # crawl.open_api plugin has to be called before.
+        if not self._is_api_spec_available():
+            om.out.information("Could not find an API specification, skip")
```
andresriancho: During a regular w3af run, this line will be printed to the output MANY times. I recommend using `om.out.debug` here.

andresriancho: Also the message could be improved: "Could not find an API specification in the KB, skipping open_api_auth"

w3af/plugins/audit/open_api_auth.py
```diff
@@ -0,0 +1,372 @@
...
+            return
+
+        # The API spec must define authentication mechanisms.
+        if not self._has_security_definitions_in_spec():
+            om.out.information("The API spec has no security definitions, skip")
```
andresriancho: Change message to `The Open API spec has no security definitions, not running open_api_auth checks.`

Also, this should be printed only once per each open-api specification document w3af finds.

w3af/plugins/audit/open_api_auth.py
```diff
@@ -0,0 +1,372 @@
...
+            return
+
+        # The check only runs if the API specification
+        # requires authentication for a endpoint.
+        if not self._has_auth(freq):
+            om.out.information("Request doesn't have auth info, skip")
```
andresriancho: Improve message and change to debug.

w3af/plugins/audit/open_api_auth.py
```diff
@@ -0,0 +1,372 @@
...
+
+    def _check_response_code(self, freq, response):
+        """
+        Check if an HTTP request contains an expected response code.
+
+        :param freq: A FuzzableRequest to the API endpoint.
+        :param response: The HTTP response associated with the fuzzable request.
+        """
+        if response.get_code() not in self._expected_codes:
+            desc = 'A %s request without auth info was send to %s. ' \
```
andresriancho: `... request without authentication information ...`

`was sent to`

`replied with unexpected HTTP response code %d (expected one of 401, 403, etc.)`

And add: `This is a strong indicator of a REST API authentication bypass`.

w3af/plugins/audit/open_api_auth.py
```diff
@@ -0,0 +1,372 @@
...
+                   'The server replied with an unexpected code (%d), expected %s' \
+                   % (freq.get_method(), freq.get_url(),
+                      response.get_code(), self._expected_codes)
+
+            # Should severity depend on response codes?
+            # For example, 2xx codes results to HIGH, but 4xx may be MEDIUM/LOW
```
andresriancho: Yes, this is a good idea to remove false positives.

The way I recommend doing it is to add a new check below `if not self._has_auth(freq)`, like this:

```python
        if not self._response_is_successful(freq):
            om.out.debug("OpenAPI request {some request info here} did not return a successful response. Skipping open_api_auth check for this endpoint.")
```

Where `_response_is_successful` would send the request (with credentials) and check that the response code is 200, 201, or some other codes which are commonly used to indicate success.

w3af/plugins/audit/open_api_auth.py
```diff
@@ -0,0 +1,372 @@
...
+        which should be called before.
+
+        The plugins use the global knowledge base to share the API spec.
+
+        :return: True if API specification is available, False otherwise.
+        """
+        if self._spec:
+            return True
+
+        specification_handler = kb.kb.raw_read('open_api', 'specification_handler')
```
andresriancho: What if there is more than one openapi document found during a scan?

Maybe you should store a list and call it `kb.kb.raw_read('open_api', 'specification_handlers')` (added an `s` at the end)

This also means that you'll have to touch other areas of this plugin to process all specification handlers, not just one.

w3af/plugins/audit/open_api_auth.py
```diff
@@ -0,0 +1,372 @@
...
+          * Global 'security' section describes authentication mechanisms
+            for operations which don't have their own 'security' section
+
+        Currently the check is pretty simple:
+          * Remove authentication info from the request (API key, header, etc).
+          * Send the modified request to the endpoint.
+          * Check if the response has 401 error code (access denied).
+
+        A couple of important notes:
+          * The plugin requires the 'crawl.open_api' plugin to be enabled.
```
andresriancho: You can force that. Plugins can have other plugins as dependencies.
https://github.com/andresriancho/w3af/blob/master/w3af/core/controllers/plugins/plugin.py#L123

andresriancho: When plugin A depends on B, and plugin A is enabled, plugin B will be automatically enabled too.

andresriancho (commented):
Awesome job!

w3af/plugins/audit/open_api_auth.py
artem-smotrakov: I also thought about it. In a perfect world, an application should return 401 if no auth info is present at all. 403 should be usually returned if a request contains auth info, but authorization failed. So now the plugin assumes that it lives in a perfect world :) I think it is not a big deal If an application returns 403 instead of 401 but strictly speaking it doesn't sound correct to me. Anyway, I don't mind adding 403 here. I would even let a user to configure it - it may be next improvement.

w3af/plugins/audit/open_api_auth.py
artem-smotrakov: Do you mean adding `_response_is_successful` check between `_remove_auth` and `_remove_auth` and `_uri_opener.send_mutant` calls in `audit()` method? But wasn't this request already sent? Is the result already available in `orig_response`?

What would be a definition of a successful response? 200 may be not enough since a endpoint may return 201, 202, etc. In fact, response codes should be defined in the spec. In general, it would be a nice improvement. Do you think such a check is strictly necessary?

In fact, I meant another thing here. I am talking about assigning different severity levels what we report a vulnerability to the KB. It should be done if/after `_response_is_successful` check passed.

w3af/plugins/audit/open_api_auth.py
artem-smotrakov: That's nice, I didn't know about it. Let me add such a dependency.

artem-smotrakov:
Here is a couple of updates:

- `audit.open_api_auth` now depends on `crawl.open_api`.
- `audit.open_api_auth` now expects multiple specs.
- Updated debug messages.
- Added `TestOpenAPIAuthMultipleSpecs`.

I also have a couple of comments/questions. Please let me know what you think.

artem-smotrakov:
Still waiting for feedback (and for other pull requests too). Please take a look when you have time.

PR #17375, opened 2018-10-31
Fix #17374 if not present install npm

sbnsec:


CustomBread:
see : https://github.com/andresriancho/w3af/issues/17374

andresriancho:
Thanks for reporting the issue and sending the PR.

I prefer to see this implemented by adding the packages to `CORE_SYSTEM_PACKAGES`, like we see here:

https://github.com/andresriancho/w3af/blob/master/w3af/core/controllers/dependency_check/platforms/ubuntu1204.py#L34

Would that make sense to you?

CustomBread:
Thanks for your answer @andresriancho ,
Well that makes sense and at the beginning it seems cleaner but :
- The npm pakage is not present on the system default repository,
  So the devs of npm are maintaining a script to ensure a clean install of the repos. this script is agnostic of the distrib/version.

As retriveJS is installed that way (in base_plateform) 
I would guess was that it would be more intuitive to install npm the same way ( as npm is the only dependency of retriveJS ) 
- Also, it can't directly install using this way (https://github.com/andresriancho/w3af/blob/master/w3af/core/controllers/dependency_check/platforms/ubuntu1204.py#L34) because the repo is not present.

If you are not agree with my point of view, What would you suggest to handle the repository problem ? 

Thanks :).
Have a nice day.

andresriancho:
At least in Ubuntu the `npm` package is there:
 * https://packages.ubuntu.com/xenial/npm (16.04)
 * https://packages.ubuntu.com/bionic/npm (18.04)

Or maybe I'm missing something? Are the versions in the ubuntu repositories too old?

CustomBread:
Okay, true, I made a new PR. https://github.com/andresriancho/w3af/pull/17407

Issue #8691, opened 2015-03-04
Find hidden gateway payment confirmation page


andresriancho:
## Contributor wanted

This issue is [easy](https://github.com/andresriancho/w3af/issues?q=is%3Aopen+is%3Aissue+label%3Aeasy) and looking for a contributor, just add a comment saying you'll do it's yours!

Not sure where to start from? Take a look at our [First steps as a contributor](https://github.com/andresriancho/w3af/wiki/First-steps-as-a-contributor) and [Contributing 101](https://github.com/andresriancho/w3af/wiki/Contributing-101) wiki pages.

Other interesting tasks which can be completed by new `w3af` contributors can be found [here](https://github.com/andresriancho/w3af/issues?q=is%3Aopen+is%3Aissue+label%3Aeasy).
## Task description

Port the [knowledge from this tool](https://github.com/wireghoul/presentations/blob/master/BHAsia2014/returnbrute.pl) into a w3af plugin. Description from the tool itself:

> Generate a list of potential confirmation pages to brute force with a tool 
> like wget/wfuzz/burp. Useful if you're pentesting a payement gateway that 
> uses a hidden payment confirmation page which is accessed by the payment 
> gateway and not intended for public access.

I'm worried about this generating requests that pykto already generates, should review the nikto db before adding all these, and also send the GET requests with `cache=True`

pedromachuca:
Hello, 
It is my first contribution, I will try to resolve this.

andresriancho:
Sounds good, go ahead. Thanks!

pedromachuca:
Hello, I created a pull request for this issue: https://github.com/andresriancho/w3af/pull/17399

I hope I did not make a mistake during the process, anyway I waiting to hear what can I do better now.

PR #17399, opened 2018-11-10
[WIP] Trying to fix issue #8691 - Find hidden gateway payment confirmation page 

pedromachuca:
This work is in progress.

I started by taking the url_fuzzer.py as an example.
I looked at the Nikto db at https://raw.github.com/sullo/nikto/master/program/databases/db_tests and
https://raw.githubusercontent.com/andresriancho/w3af/master/w3af/plugins/crawl/pykto/scan_database.db

It seems that there are some directories and files in common but most of them doesn't appear in the nikto db (assuming I looked at the right ones). Here are the common elements : 

Directories :  
/inc/
/include/
/includes/
/lib/
/module/
/modules/
/shop/
/store/
/servlet/
/cgi/
/cgi-bin/
/cgibin/

Files :
payment
payments

I tested my plugin with the python SimpleHTTPServer, I created directories and files corresponding to the one my plugin is generating.
w3af was able to find the directories and files created.

This is my first contribution I'm eager to ear some advices or directions I should continue toward in order to merge this plugin.

I hope I explained myself well enough.

### DIFF
--- a/w3af/plugins/crawl/url_payment_finder.py
+++ b/w3af/plugins/crawl/url_payment_finder.py
@@ -1,5 +1,5 @@
 """
-url_payment_fuzzer.py
+url_payment_finder.py
 
 Copyright 2006 Andres Riancho
 
--- a/w3af/plugins/crawl/url_payment_finder.py
+++ b/w3af/plugins/crawl/url_payment_finder.py
@@ -1,0 +1,218 @@
+"""
+url_payment_fuzzer.py
+
+Copyright 2006 Andres Riancho
+
+This file is part of w3af, http://w3af.org/ .
+
+w3af is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation version 2 of the License.
+
+w3af is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with w3af; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+"""
+from itertools import chain, repeat, izip
+
+import w3af.core.controllers.output_manager as om
+import w3af.core.data.kb.knowledge_base as kb
+from w3af.core.controllers.plugins.crawl_plugin import CrawlPlugin
+from w3af.core.controllers.core_helpers.fingerprint_404 import is_404
+from w3af.core.data.dc.headers import Headers
+from w3af.core.data.bloomfilter.scalable_bloom import ScalableBloomFilter
+from w3af.core.data.parsers.doc.url import URL
+from w3af.core.data.kb.info import Info
+from w3af.core.data.request.fuzzable_request import FuzzableRequest
+
+
+class url_payment_fuzzer(CrawlPlugin):
+    """
+    Try to find hidden gateway payment confirmation page.
+    :author: Coiffey Pierre (pierre.coiffey@gmail.com)
+    """
+    _dirs = ('/', '/inc/', '/include/', '/include/pay/', '/includes/', '/includes/pay/', '/lib/',
+             '/libraries/', '/module/', '/module/pay/', '/modules/', '/modules/pay/', '/payment/',
+             '/shop/', '/store/', '/svc/', '/servlet/', '/cgi/', '/cgi-bin/', '/cgibin/',
+    )
+
+    _files = ('pay', 'payment', 'success', 'paymentsuccess', 'paymentcomplete', 'paymentsuccessful',
+              'successful', 'paid', 'return', 'valid', 'validpay', 'validate', 'validatepayment',
+              'validatepay', 'validation', 'complete', 'completepay', 'completepayment',
+              'trxcomplete', 'transactioncomplete', 'final', 'finished',
+    )
+    _exts = ('', '.php', '.asp', '.aspx', '.jsp', '.py', 
+             '.pl', '.rb', '.cgi', '.php3',  '.php4', '.php5',
+    )
+
+    def __init__(self):
+
+        CrawlPlugin.__init__(self)
+
+        self._headers = None
+        self._first_time = True
+        self._fuzz_images = False
+        self._seen = ScalableBloomFilter()
+
+    def crawl(self, fuzzable_request):
+        """
+        Searches for new Url's using fuzzing.
+
+        :param fuzzable_request: A fuzzable_request instance that contains
+                                    (among other things) the URL to test.
+        """
+        url = fuzzable_request.get_url()
+        self._headers = Headers([('Referer', url.url_string)])
+
+        if self._first_time:
+            self._verify_head_enabled(url)
+            self._first_time = False
+
+        # First we need to delete fragments and query strings from URL.
+        url = url.uri2url()
+
+        # And we mark this one as a "do not return" URL, because the
+        # core already found it using another technique.
+        self._seen.add(url)
+
+        self._verify_head_enabled(url)
+
+        if self._head_enabled():
+            response = self._uri_opener.HEAD(url, cache=True,
+                                             headers=self._headers)
+        else:
+            response = self._uri_opener.GET(url, cache=True,
+                                            headers=self._headers)
+
+        if response.is_text_or_html():
+            mutants_chain = chain(self._mutate_path(url))
+
+            url_repeater = repeat(url)
+            args = izip(url_repeater, mutants_chain)
+
+            self.worker_pool.map_multi_args(self._do_request, args)
+
+    def _do_request(self, url, mutant):
+        """
+        Perform a simple GET to see if the result is an error or not, and then
+        run the actual fuzzing.
+        """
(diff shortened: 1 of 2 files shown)


w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
+"""
+url_payment_finder.py
```
andresriancho: Maybe rename it to `payment_webhook_finder.py`?

All crawl plugins find urls, so it is a little bit too generic to use `url_...` naming.

andresriancho: Remember to rename the filename, this comment content and class name.

w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
...
+from w3af.core.data.dc.headers import Headers
+from w3af.core.data.bloomfilter.scalable_bloom import ScalableBloomFilter
+from w3af.core.data.parsers.doc.url import URL
+from w3af.core.data.kb.info import Info
+from w3af.core.data.request.fuzzable_request import FuzzableRequest
+
+
+class url_payment_finder(CrawlPlugin):
+    """
+    Try to find hidden gateway payment confirmation page.
```
andresriancho: I believe a better description would be:

```
Find hidden payment gateway webhooks.
```

w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
...
+             '.pl', '.rb', '.cgi', '.php3',  '.php4', '.php5',
+    )
+
+    def __init__(self):
+
+        CrawlPlugin.__init__(self)
+
+        self._headers = None
+        self._first_time = True
+        self._fuzz_images = False
```
andresriancho: This is not used, remove.

w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
...
+        Searches for new Url's using fuzzing.
+
+        :param fuzzable_request: A fuzzable_request instance that contains
+                                    (among other things) the URL to test.
+        """
+        url = fuzzable_request.get_url()
+        self._headers = Headers([('Referer', url.url_string)])
+
+        if self._first_time:
+            self._verify_head_enabled(url)
```
andresriancho: The goal of sending HTTP HEAD requests in pykto was to reduce the network traffic. This plugin doesn't send that many HTTP requests, so it is fine to just send GET requests. You can go ahead and remove all the logic associated with HEAD.

The good thing about this HEAD code is that it made me think about HTTP methods! Some / most payment gateway webhooks will receive data over POST. Some web application frameworks will only route the traffic to the controller if the method (POST) is used.

So I believe that for each URL we need to try GET and POST.

w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
...
+            self._first_time = False
+
+        # First we need to delete fragments and query strings from URL.
+        url = url.uri2url()
+
+        # And we mark this one as a "do not return" URL, because the
+        # core already found it using another technique.
+        self._seen.add(url)
```
andresriancho: This doesn't look OK.

If the framework calls this plugin with `http://host/path?id=1` and `http://host/path?id=2` then the plugin will send duplicated HTTP requests to:

 * `http://host/pay`
 * `http://host/payment`
 * `http://host/success`
 * ...

Right?

What we want to do is:
 * Transform the input URI to a URL
 * Check if we already performed the analysis on that URL. If we did, then return
 * Perform the analysis on the URL, because it was never done before

w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
...
+
+    def crawl(self, fuzzable_request):
+        """
+        Searches for new Url's using fuzzing.
+
+        :param fuzzable_request: A fuzzable_request instance that contains
+                                    (among other things) the URL to test.
+        """
+        url = fuzzable_request.get_url()
+        self._headers = Headers([('Referer', url.url_string)])
```
andresriancho: Do not use instance attributes to store this. Take into account that w3af uses multiple threads, if you store the referer header in an instance attribute, one thread will override the value of the other thread.

w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
...
+        if (response.get_code() in (403, 401) or not(is_404(response))):
+
+            # Create the fuzzable request and send it to the core
+            fr = FuzzableRequest.from_http_response(response)
+            self.output_queue.put(fr)
+            
+            #
+            #   Save it to the kb (if new)!
+            #
+            if response.get_url() not in self._seen and response.get_url().get_file_name():
```
andresriancho: Here you use `self._seen` for a completely different thing. Do not use "the same DB table" to store different object types.

w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
...
+        
+        if 'HEAD' in allowed_methods:
+            self._head = True
+        else:
+            self._head = False
+
+    def _head_enabled(self):
+        return self._head
+
+    def get_plugin_deps(self):
```
andresriancho: We don't need this dependency because HEAD will not be used, just remove the whole method.

w3af/plugins/crawl/url_payment_finder.py
```diff
@@ -0,0 +1,219 @@
...
+        current one.
+        """
+        return ['infrastructure.allowed_methods']
+
+    def get_long_desc(self):
+        """
+        :return: A DETAILED description of the plugin functions and features.
+        """
+        return """
+        This plugin will try to find new URL's based on the input. If the input
```
andresriancho: `This plugin will try to find new payment gateway webhook URLs based on the input`

andresriancho (changes requested):
Excellent first step! Please apply the requested changes so that we can merge it.

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: Thank you for the detailed description for the requested changes. It's really helpfull.
I started with the simple changes like this one. I will try make the other changes before next week so I may come back with some questions. I will comment "done" for the ones I've done already.

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: Done

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: Done

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: Done

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: Done

w3af/plugins/crawl/url_payment_finder.py
andresriancho: Still an issue

w3af/plugins/crawl/url_payment_finder.py
andresriancho: Still an issue

w3af/plugins/crawl/url_payment_finder.py
andresriancho: Still an issue

w3af/plugins/crawl/url_payment_finder.py
andresriancho: Still an issue

w3af/plugins/crawl/url_payment_finder.py
andresriancho: Method was removed but is called in the plugin.

andresriancho:
@pedromachuca is this PR ready for a final review? Should I give it a try?

pedromachuca (commented):
Hello, 
I didn't commit yet because I did not modify the last issue yet. 
However I made all the other changes.

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: For this one I took the url_payment as an example and I'm not sure how to modify it.

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: This should be ok.

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: Yes I agree, the modifications should be ok.

w3af/plugins/crawl/url_payment_finder.py
pedromachuca: I removed all the logic associated with HEAD. I did not try for the POST request though.

andresriancho:
Performed a PR take over and completed the task myself in https://github.com/andresriancho/w3af/commit/528b84b539359b3d438115932552057b92bc2f5a

The plugin was sending **a lot of HTTP requests** so I decided to implement *_get_extensions_for_fuzzing* to reduce that amount.

Also decided to add `POST` as a method for the test

PR #17407, opened 2018-11-14
Fix #17374 Install npm Automaticaly

sbnsec:
I think npm should be installed automatically as it is used to install retireJS. 
I made a new PR after the advises from @andresriancho, Thanks to him :).


CustomBread:
Yeah !

Issue #16271, opened 2017-12-04
Improve url_fuzzer plugin using tricks from bfac


andresriancho:
There is a nice tool called [bfac](https://github.com/mazen160/bfac) which has some url fuzzing tricks which are not implemented in [url_fuzzer](https://github.com/andresriancho/w3af/blob/a53d59b73c493013a90fccf468cf443e47f869d0/w3af/plugins/crawl/url_fuzzer.py).

It would be nice to get those url fuzzing tricks implemented into our url_fuzzer, giving credit in the source code to @mazen160 for his ideas on how to create the interesting files.

andresriancho:
@alexandreborgo sure! Please go ahead!

I'm a little bit worried about the overlap between things that `bfac` does and other w3af plugins. For example, `find_dvcs.py` does search for `.gitignore`. So that one shouldn't be added to `url_fuzzer`.

`dot_ds_store.py` does find `.DS_Store` files, which are also mentioned in `bfac`.

Other extensions and transformations might have the same problem. I recommend doing an in-depth analysis of each path transformation mentioned here: https://github.com/mazen160/bfac/blob/master/bfac#L568-L726 and compare it with what w3af has.

For example, I found out that  there was an overlap with DSStore by running:
```
find . -type f -print0 | xargs -0 grep DS_Store
```

And inspecting the output.

Could you do that analysis and after that write the code? Please write the analysis results here in a comment.

alexandreborgo:
Yes you're right, I hadn't thought about overlapping with other plugins.

Doing an analysis:
- bfac backup_testing_level1, backup_testing_level2 and backup_testing_level3: 
 	- change/add extension (backup, copy, save, etc), some are find by `url_fuzzer.py`, few aren't
 	- file transformation (inserting `Copy_of_` for example), no result on w3af
- bfac backup_testing_level4:
 	- files related to archive and compression, most are find by `url_fuzzer.py`, few aren't
 	- `bfac` looks for `site + file_path + '.'` and `site + file_path + '::$DATA'`
- bfac backup_testing_level5: 
	- files related to VCS, `find_dvcs.py` can find
	- files related to IDE setting `.idea/*`, seems like `pykto.py` looks for it
	- files related to composer, same as IDE setting

Note: some other plugins like `dir_file_bruter` and `pykto` try to find common files with these extensions, for examples `index.html.zip`, `.htaccess.bak`.

So `url_fuzzer` shouldn't look for common files, other plugins already take care of it.

[...]

PR #17434, opened 2018-11-23
Implement bfac tricks into url_fuzzer

alexandreborgo:
See #16271 

Hey, here is a first try.

What I added:
- new `appendables`, `backup_exts` and `file_types`
- new set of prependables (like `Copy_of_`)
- a function `_mutate_by_prepending` to generate new URLs from this set
- a function `_mutate_file_name` to generate change on the name of the file like : `index.html` -> `index-backup.html`, `index (copy).html`, `.index.html.swp`, etc
- and ofc the calls to the new functions

I checked that new patterns don't create duplicates of existing URLs.
I don't know if the function `_mutate_file_name` is the best way to do it. I used the same technique as `bfac`, what do you think about this function ?

I performed some tests, everything works properly but I found something strange. The plugin finds all 26 URLs it could find on the server but the list of fuzzable requests contains several times `http://localhost/`.
See the attached file [output.txt](https://github.com/andresriancho/w3af/files/2609461/output.txt) for the list of file on the test server and w3af output.

This bug as something to do with the new function `_mutate_file_name`. It found 6 new URLs and there's 6 more `http://localhost/` in the fuzzable requests. If the plugin don't call the function the result is good. I'm sure that the function doesn't create this URL directly. I'll work on it.

### DIFF
--- a/w3af/plugins/crawl/url_fuzzer.py
+++ b/w3af/plugins/crawl/url_fuzzer.py
@@ -53,6 +53,9 @@ class url_fuzzer(CrawlPlugin):
                     '.sql.gz', '.bak.sql', '.bak.sql.gz', '.bak.sql.bz2',
                     '.bak.sql.tar.gz'
                    )
+    _prependables = ('_', '.', '~', '.~', 'Copy_', 'Copy_of_', 'Copy_(1)_of_',
+                     'Copy_(2)_of_', 'Copy ', 'Copy of ', 'backup-'
+                    )
     _backup_exts = ('tar.gz', '7z', 'gz', 'cab', 'tgz', 'gzip',
                     'bzip2', 'zip', 'rar'
                     )
--- a/w3af/plugins/crawl/url_fuzzer.py
+++ b/w3af/plugins/crawl/url_fuzzer.py
@@ -311,6 +311,33 @@ class url_fuzzer(CrawlPlugin):
                 url_copy.set_extension(filetype)
                 yield url_copy
 
+    def _mutate_file_name(self, url):
+        filename = url.get_file_name()
+        if filename:
+            domain = url.get_domain_path()
+            url_string = domain.url_string
+            name = filename[:filename.rfind(u'.')]
+            extension = url.get_extension()
+
+            mutate_name_testing = (
+                url_string + '#' + filename + '#',
+                url_string + name + ' (copy).' + extension,
+                url_string + name + ' - Copy.' + extension,
+                url_string + name + ' copy.' + extension,
+                url_string + '.~lock.' + filename + '#',
+                url_string + name + '-backup.' + extension,
+                url_string + name + '-bkp.' + extension,
+                url_string + '.' + name + '.swp',
+                url_string + '_' + name + '.swp',
+                url_string + '.' + name + '.swo',
+                url_string + '_' + name + '.swo',
+                url_string + '~' + name + '.tmp'
+            )
+
+            for change in mutate_name_testing:
+                newurl = URL(change)
+                yield newurl
+
     def _mutate_path(self, url):
         """
         Mutate the path instead of the file.
--- a/w3af/plugins/crawl/url_fuzzer.py
+++ b/w3af/plugins/crawl/url_fuzzer.py
@@ -319,17 +319,22 @@ class url_fuzzer(CrawlPlugin):
         if filename:
             domain = url.get_domain_path()
             url_string = domain.url_string
-            name = filename[:filename.rfind(u'.')]
             extension = url.get_extension()
+
+            if extension:
+                extension = '.' + extension
+                name = filename[:filename.rfind(u'.')]
+            else:
+                name = filename
 
             mutate_name_testing = {
                 url_string + '#' + filename + '#',
-                url_string + name + ' (copy).' + extension,
-                url_string + name + ' - Copy.' + extension,
-                url_string + name + ' copy.' + extension,
+                url_string + name + ' (copy)' + extension,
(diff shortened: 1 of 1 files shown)


w3af/plugins/crawl/url_fuzzer.py
```diff
@@ -41,19 +41,30 @@ class url_fuzzer(CrawlPlugin):
     Try to find backups, and other related files.
     :author: Andres Riancho (andres.riancho@gmail.com)
     """
-    _appendables = ('~', '.tar.gz', '.gz', '.7z', '.cab', '.tgz',
-                    '.gzip', '.bzip2', '.inc', '.zip', '.rar', '.jar', '.java',
-                    '.class', '.properties', '.bak', '.bak1', '.bkp', '.back',
-                    '.backup', '.backup1', '.old', '.old1', '.$$$'
+    _appendables = ('~', '~~', '_', '.', '.tar.gz', '.gz', '.7z', '.cab',
```
andresriancho: Now that these lists are getting larger, we have the risk of duplicating one item in them.

Let's change the type from tuple (current) to set (desired). By doing so we make sure that even if we add duplicates in the variable definition, the set will eliminate them.

Just change: `_appendables = (...)` with `_appendables = {...}`.

The same change should be applied to all other class attributes that define lists like this in the url_fuzzer class.

w3af/plugins/crawl/url_fuzzer.py
```diff
@@ -269,6 +314,33 @@ def _mutate_file_type(self, url):
                 url_copy.set_extension(filetype)
                 yield url_copy
 
+    def _mutate_file_name(self, url):
+        filename = url.get_file_name()
+        if filename:
+            domain = url.get_domain_path()
+            url_string = domain.url_string
+            name = filename[:filename.rfind(u'.')]
```
andresriancho: What happens here when the input URL is `http://w3af.org/foobar`  (there is no point in the filename)

w3af/plugins/crawl/url_fuzzer.py
```diff
@@ -269,6 +314,33 @@ def _mutate_file_type(self, url):
...
+            extension = url.get_extension()
+
+            mutate_name_testing = (
+                url_string + '#' + filename + '#',
+                url_string + name + ' (copy).' + extension,
```
andresriancho: And also, following the same example above where the filename has no "name", what happens in these lines when `extension` is empty? Does it make sense to send these tests?

These cases should be investigated and handled:
 * http://w3af.org/foobar
 * http://w3af.org/foobar.
 * http://w3af.org/.hidden
 * http://w3af.org/.

andresriancho (changes requested):

w3af/plugins/crawl/url_fuzzer.py
```diff
@@ -269,6 +314,33 @@ def _mutate_file_type(self, url):
...
+                url_string + name + '-bkp.' + extension,
+                url_string + '.' + filename + '.swp',
+                url_string + '_' + filename + '.swp',
+                url_string + '.' + filename + '.swo',
+                url_string + '_' + filename + '.swo',
+                url_string + '~' + filename + '.tmp'
+            )
+
+            for change in mutate_name_testing:
+                newurl = URL(change)
```
andresriancho: Just in case, do a try/except around the URL instance creation, we never know what could go wrong when crafting these URLs.

andresriancho:
I took a quick look at the output.txt file and noticed these lines:

```
The list of fuzzable requests is:
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
- Method: GET | http://localhost/
```

This is what you said was strange, right?

I also believe it is strange... but can't understand why it is like that...

Could you share the w3af script you are using to run the scan, and a little bit more about the target web application? Thanks!

w3af/plugins/crawl/url_fuzzer.py
alexandreborgo: Done.

w3af/plugins/crawl/url_fuzzer.py
alexandreborgo: Yes it wasn't handling these files correctly, I did some modifications and I believe it does it well now.

Here are the tests I did and the results : https://pastebin.com/WG5szsnB (GitHub importing files system doesn't work right now so I put the output on Pastebin)

w3af/plugins/crawl/url_fuzzer.py
alexandreborgo: See below's answer.

w3af/plugins/crawl/url_fuzzer.py
alexandreborgo: I'm not sure if you want the try/except in w3af or if it is just to find out what's generating all `http://localhost/` in the list fuzzable requests.

I did this: 
```python
try :
     newurl = URL(change)
     yield newurl
except ValueError as exception:
     om.out.information('Error while generating a new URL: '\
     + exception)
```

And nothing wrong came from it during all my tests.

Let me know if I have to add the exception to the pull request.

andresriancho: Yes, the code you show in the comment is what I wanted, just in case... you never know...

The only change I would propose to that code is to move the `yield newurl` to the `else:` section.

Issue #17026, opened 2018-05-24
Identify Server-Side Template Injection vulnerabilities


andresriancho:
https://portswigger.net/blog/server-side-template-injection

Issue #12708, opened 2015-09-22
Ruby YAML deserialization bug


andresriancho:
Detect Ruby's YAML deserialization bug:

```text
string = "--- !ruby/hash:Usern" +
"email: richard@example.com"

YAML::load(string)
 => #<User id: 1, email: "richard@example.com">
```

http://www.sitepoint.com/anatomy-of-an-exploit-an-in-depth-look-at-the-rails-yaml-vulnerability/

Issue #9351, opened 2015-04-07
New plugin: Content type incorrectly stated


andresriancho:
## Context
I've seen this in some burp reports:

>Summary
The response contains the following Content-type statement:
`Content-Type: text/html; charset=utf-8`

> The response states that it contains HTML. However, it actually appears to contain JSON.

> Issue background
If a web response specifies an incorrect content type, then browsers may process the response in unexpected ways. If the specified content type is a renderable text-based format, then the browser will usually attempt to parse and render the response in that format. If the specified type is an image format, then the browser will usually detect the anomaly and will analyze the actual content and attempt to determine its MIME type. Either case can lead to unexpected results, and if the content contains any user-controllable data may lead to cross-site scripting or other client-side vulnerabilities.  In most cases, the presence of an incorrect content type statement does not constitute a security flaw, particularly if the response contains static content. You should review the contents of the response and the context in which it appears to determine whether any vulnerability exists.  

> Issue remediation
For every response containing a message body, the application should include a single Content-type header which correctly and unambiguously states the MIME type of the content in the response body.  

## Tasks
 - [ ] Create a new plugin that will detect this vulnerability
 - [ ] Create unittests
 - [ ] Add the vulnerability to vulndb/data
 - [ ] Create a new python-sdk release for the new vulndb/data
 - [ ] Update requirements.py to point to the new python-sdk release
 - [ ] Link vulns.py to the new vulnerability ID from vulndb/data

## Algorithm
TBD

Issue #8662, opened 2015-03-03
Research: Detect hash extension attacks


andresriancho:
Is it possible to use https://github.com/bwall/HashPump to automate the detection of hash length extension attacks?

PS: Please note the python bindings

PR #17549, opened 2019-01-15
Support for Expect-CT header

mleblebici:
Added checks for Expect-CT header


w3af/plugins/grep/expect_ct.py
```diff
@@ -0,0 +1,94 @@
...
+
+    def grep(self, request, response):
+        """
+        Check if HTTPS responses have the Expect-CT header set.
+
+        :param request: The HTTP request object.
+        :param response: The HTTP response object
+        :return: None, all results are saved in the kb.
+        """
+        if self._reports > MAX_REPORTS:
```
andresriancho: This check might not be required, see https://github.com/andresriancho/w3af/blob/3a1b04fd5a519f6e981d714350b234061cdc38e9/w3af/core/data/kb/info_set.py#L85

andresriancho (changes requested):
This is one of the best PRs I've ever received. It has unittests, adds a new vuln check, and you even modified related parts of the framework (vulns.py and strange_headers.py) to make the whole thing work :clap: :clap:

w3af/plugins/grep/expect_ct.py
```diff
@@ -0,0 +1,90 @@
...
+        """
+
+        if request.get_url().get_protocol() != 'https':
+            return
+
+        expect_ct_header_value, _ = response.get_headers().iget(ECT_HEADER, None)
+        if expect_ct_header_value is not None:
+            return
+
+        self._reports += 1
```
andresriancho: You know what? I made a mistake with the previous comment on `self._reports`. I was thinking only about functionality: InfoSet will take care of the max reports, so `self._reports` is not required on that side.... **but if we think about performance** then the story is different! By using `self._reports` you're preventing a lot of processing after the 50th reported vulnerability, which is great!

Please re-add the self._reports check at the beginning of this method.

Sorry for this

andresriancho (changes requested):

andresriancho:
giphy

andresriancho:
Thanks for your PR! Maybe you're interested in contributing with other things? Some ideas here:

 * https://github.com/andresriancho/w3af/issues/8662
 * https://github.com/andresriancho/w3af/issues/9351
 * https://github.com/andresriancho/w3af/issues/17026
 * https://github.com/andresriancho/w3af/issues/12708

Or anything else that is on your radar. Thanks!

PR #17668, opened 2019-03-03
Created Endpoint to get current state details of all found vulnerabilities in running state and status

krtkvrm:
**Current Problem:**
- `/scans/:id/kb` returns list of all **found** vulnerabilities
- In order to get details for all found vulnerabilities,  we need to use api request to `/scans/:id/kb` for all `ids`.

**Solution:**
Created Endpoint `/scans/:id/state` that will return detailed kb and status.

### DIFF
--- a/w3af/core/ui/api/resources/state.py
+++ b/w3af/core/ui/api/resources/state.py
@@ -42,7 +42,6 @@ def get_current_state(scan_id):
         finding = finding_to_json(finding, scan_id, finding_id, detailed=True)
         kb_list.append(finding)
     result['kb'] = kb_list
-    scan_info = get_scan_info_from_id(scan_id)
     if scan_info is None:
         abort(404, 'Scan not found')
     exc = scan_info.exception
--- a/w3af/core/ui/api/resources/state.py
+++ b/w3af/core/ui/api/resources/state.py
@@ -1,0 +1,54 @@
+"""
+state.py
+
+Copyright 2019 Kartik Verma
+
+This file is part of w3af, http://w3af.org/ .
+
+w3af is free software; you can redistribute it and/or modify
+it under the terms of the GNU General Public License as published by
+the Free Software Foundation version 2 of the License.
+
+w3af is distributed in the hope that it will be useful,
+but WITHOUT ANY WARRANTY; without even the implied warranty of
+MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+GNU General Public License for more details.
+
+You should have received a copy of the GNU General Public License
+along with w3af; if not, write to the Free Software
+Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
+
+"""
+from w3af.core.ui.api import app
+from w3af.core.ui.api.utils.auth import requires_auth
+from w3af.core.ui.api.utils.scans import get_scan_info_from_id
+from w3af.core.ui.api.resources.kb import finding_to_json
+from w3af.core.ui.api.utils.error import abort
+import w3af.core.data.kb.knowledge_base as kb
+
+from flask import jsonify
+
+
+@app.route('/scans/<int:scan_id>/state/', methods=['GET'])
+@requires_auth
+def get_current_state(scan_id):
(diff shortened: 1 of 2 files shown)


w3af/core/ui/api/resources/state.py
```diff
@@ -0,0 +1,54 @@
...
+    if scan_info is None:
+        abort(404, 'Scan not found')
+    kb_list = []
+    result = {}
+    result['id'] = scan_id
+    for finding_id, finding in enumerate(kb.kb.get_all_findings()):
+        finding = finding_to_json(finding, scan_id, finding_id, detailed=True)
+        kb_list.append(finding)
+    result['kb'] = kb_list
+    scan_info = get_scan_info_from_id(scan_id)
```
andresriancho: You already got this info at the beginning of the method

andresriancho:
The endpoint for [/scans/:id/kb](https://github.com/andresriancho/w3af/blob/master/w3af/core/ui/api/resources/kb.py#L32) returns a list of all vulnerabilities found in a scan.

The vulnerability list is generated using:

```python
    for finding_id, finding in enumerate(kb.kb.get_all_findings()):
        if matches_filter(finding, request):
            data.append(finding_to_json(finding, scan_id, finding_id))
```

The change you propose uses:

```python
    for finding_id, finding in enumerate(kb.kb.get_all_findings()):
        finding = finding_to_json(finding, scan_id, finding_id, detailed=True)
        kb_list.append(finding)
```

Where there is no filter and `detailed=True` was added. The change you propose also adds some scan data (exceptions, status, etc.) to the result.

What I understand from your change is that (because of performance issues) you don't want your client to do this:

```
vulns_without_details = client.get('/scans/:id/kb')
for vuln_without_detail in vulns_without_details:
    vuln_with_detail = client.get('/scans/<int:scan_id>/kb/<int:vulnerability_id>')
    process(vuln_with_detail)
```

The [/scans/<int:scan_id>/kb/<int:vulnerability_id>](https://github.com/andresriancho/w3af/blob/master/w3af/core/ui/api/resources/kb.py#L69) endpoint returns the details you need, but instead of calling it N times to get all the info, you want to call the new endpoint only once.

Is my understanding correct?

andresriancho:
If my understanding of your code and problem is correct, I recommend this instead:

In [this section](https://github.com/andresriancho/w3af/blob/master/w3af/core/ui/api/resources/kb.py#L64):

```python
data.append(finding_to_json(finding, scan_id, finding_id))
```

Add a query string parameter called `detailed` (default value should be `False`) which allows the user to specify if he wants a detailed KB vuln data to be returned or not.

The parameter should be properly documented in the method doc.

andresriancho:
Please close this PR and send a new one if all this makes sense to you :+1:

andresriancho:
@vkartik97 ?

vkartik97:
@andresriancho 
1. >The /scans/int:scan_id/kb/int:vulnerability_id endpoint returns the details you need, but instead of 
    calling it N times to get all the info, you want to call the new endpoint only once.
   >
   > Is my understanding correct?

    Yes.

2. >Add a query string parameter called detailed (default value should be False) which allows the user to specify if he wants a detailed KB vuln data to be returned or not.

    Do you mean to say something like `/scans/:id/kb?detailed=True`?

3. >If my understanding of your code and problem is correct, I recommend this instead:
    >
    >In this section:
    >
    >           `data.append(finding_to_json(finding, scan_id, finding_id))`
    >Add a query string parameter called detailed (default value should be False) which allows the user to specify if he wants a detailed KB vuln data to be returned or not.

     Do you mean to say if only if `/scans/:id/kb?detailed=True`, then this:
     ```python
        for finding_id, finding in enumerate(kb.kb.get_all_findings()):
             finding = finding_to_json(finding, scan_id, finding_id, detailed=True)
             kb_list.append(finding)
      ```
     else the existing `/scans/:id/kb`

vkartik97:
If these are the changes you are proposing, I can open a new PR for these

andresriancho:
Yes, I believe that is about it!

vkartik97:
@andresriancho 
Can I incorporate changes in this PR?

Issue #17672, opened 2019-03-07
mangle::sed doesn't store expressions as strings


Rakurai:
$ ./w3af_console 
w3af>>> plugins
w3af/plugins>>> mangle config sed
w3af/plugins/mangle/config:sed>>> set expressions qh/User/NotLuser/
w3af/plugins/mangle/config:sed>>> view
|------------------------------------------------------------------------------------------------------------------------|
| Setting             | Value                  | Modified   | Description                                                |
|------------------------------------------------------------------------------------------------------------------------|
| expressions         | qh/User/NotLuser/      | Yes        | Stream edition expressions                                 |
| fix_content_len     | True                   |            | Fix the content length header after mangling               |
|------------------------------------------------------------------------------------------------------------------------|
w3af/plugins/mangle/config:sed>>> save
The configuration has been saved.
w3af/plugins/mangle/config:sed>>> view
|-------------------------------------------------------------------------------------------------------------------------|
| Setting           | Value                              | Modified  | Description                                        |
|-------------------------------------------------------------------------------------------------------------------------|
| expressions       | ('q', 'h', 'User', 'NotLuser')     |           | Stream edition expressions                         |
| fix_content_len   | True                               |           | Fix the content length header after mangling       |
|-------------------------------------------------------------------------------------------------------------------------|
w3af/plugins/mangle/config:sed>>> save
Traceback (most recent call last):
  File "/Users/jason/repos/w3af/w3af/core/ui/console/console_ui.py", line 230, in _handleKey

    self._handlers[key]()
  File "/Users/jason/repos/w3af/w3af/core/ui/console/console_ui.py", line 319, in _onEnter
    self._execute()
  File "/Users/jason/repos/w3af/w3af/core/ui/console/console_ui.py", line 288, in _execute
    menu = self._context.execute(params)
  File "/Users/jason/repos/w3af/w3af/core/ui/console/menu.py", line 173, in execute
    return handler(params)
  File "/Users/jason/repos/w3af/w3af/core/ui/console/config.py", line 146, in _cmd_save
    self._configurable.set_options(self._options)
  File "/Users/jason/repos/w3af/w3af/plugins/mangle/sed.py", line 119, in set_options
    self._expressions = ','.join(option_list['expressions'].get_value())
TypeError: sequence item 0: expected string, tuple found

PR #17673, opened 2019-03-07
Separated double-purposed self._expressions variable, now it simply h…

Rakurai:
Separated double-purposed self._expressions variable, now it simply holds the comma-separated expression strings while found_expressions within the set_options() scope holds the list of match tuples parsed be re.  Fix for issue #17672.


andresriancho:
giphy

PR #17707, opened 2019-03-23
Better support for path parameters in OpenAPI plugin

mpoindexter:
Right now, the OpenAPI plugin just forces on url_parts fuzzing.  This has two disadvantages:
1.)  It's slow. Adding injections to non-parameterized parts of the routes is useless for many REST APIs
2.) It's incorrect. Path parameters don't have to use a full segment.  For example `/foo/bar/product-{param}`, the route won't be matched if there's no `product-` at the beginning of the last segment.

This PR adds a forced URL parts to fuzzable requests that indicate path segments where injection is possible. This allows the OpenAPI plugin to properly handle path parameters.

### DIFF
--- a/w3af/core/data/fuzzer/mutants/tests/test_urlparts_mutant.py
+++ b/w3af/core/data/fuzzer/mutants/tests/test_urlparts_mutant.py
@@ -129,3 +129,22 @@ class TestURLPartsMutant(unittest.TestCase):
         generated_urls = set([m.get_url().url_string for m in generated_mutants])
 
         self.assertEqual(set(expected_urls), generated_urls)
+
+    def test_forced_url_parts(self):
+        freq = FuzzableRequest(URL('http://www.w3af.com/static/foo/bar.ext'))
+        freq.set_force_fuzzing_url_parts([
+            ('/static/', False), ('foo', True), ('/bar.', False), ('ext', True)
+        ])
+
+        generated_mutants = URLPartsMutant.create_mutants(
+            freq, self.payloads, [],
+            False, self.fuzzer_config)
+
+        expected_urls = ['http://www.w3af.com/static/abc/bar.ext',
+                         'http://www.w3af.com/static/def/bar.ext',
+                         'http://www.w3af.com/static/foo/bar.abc',
+                         'http://www.w3af.com/static/foo/bar.def']
+
+        generated_urls = set([m.get_url().url_string for m in generated_mutants])
+
+        self.assertEqual(set(expected_urls), generated_urls)
--- a/w3af/core/data/fuzzer/mutants/tests/test_urlparts_mutant.py
+++ b/w3af/core/data/fuzzer/mutants/tests/test_urlparts_mutant.py
@@ -148,3 +148,22 @@ class TestURLPartsMutant(unittest.TestCase):
         generated_urls = set([m.get_url().url_string for m in generated_mutants])
 
         self.assertEqual(set(expected_urls), generated_urls)
+
+    def test_forced_url_parts_qs(self):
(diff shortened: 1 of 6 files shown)


andresriancho:
@artem-smotrakov what do you think about this PR?

w3af/plugins/crawl/open_api.py
```diff
@@ -458,4 +469,9 @@ def get_long_desc(self):
         During parsing an Open API specification, the plugin looks for parameters
         which are passed to endpoints via HTTP headers, and enables them for further testing.
         This behavior may be disabled by setting 'discover_fuzzable_headers' configuration parameter to False.
+
+        By default, URLs discovered by this plugin allow for injection at path parameter
```
andresriancho: Not sure if the description is needed here, since IMHO if this feature is implemented it will become the default and people that want to a) understand how openapi fuzzing works, b) disable this feature can go to the parameter documentation.

w3af/plugins/crawl/open_api.py
```diff
@@ -411,6 +413,14 @@ def get_options(self):
         o = opt_factory('discover_fuzzable_headers', self._discover_fuzzable_headers, d, BOOL, help=h)
         ol.add(o)
 
+        d = 'Automatic path parameter discovery for further testing'
+        h = ('By default, URLs discovered by this plugin allow for injection at path parameter'
```
andresriancho: The description needs to be much better. It is almost impossible to understand what this parameter does without looking at the code.

w3af/plugins/crawl/open_api.py
```diff
@@ -76,6 +76,7 @@ def __init__(self):
         self._no_spec_validation = False
         self._custom_spec_location = ''
         self._discover_fuzzable_headers = True
+        self._discover_fuzzable_url_parts = True
```
andresriancho: I agree with the default to True.

w3af/core/data/fuzzer/mutants/tests/test_urlparts_mutant.py
```diff
@@ -129,3 +129,22 @@ def test_valid_results_double_encoding(self):
         generated_urls = set([m.get_url().url_string for m in generated_mutants])
 
         self.assertEqual(set(expected_urls), generated_urls)
+
+    def test_forced_url_parts(self):
```
andresriancho: Would like to see a unittest which includes a query string containing the string `foo`. The expected result for that test would be that the QS is not fuzzed.

andresriancho (changes requested):

w3af/plugins/crawl/open_api.py
mpoindexter: OK, I'll take stab at improving this. I largely based the description on the discover_fuzzable_headers setting, but I agree it could be more clear.

w3af/core/data/fuzzer/mutants/tests/test_urlparts_mutant.py
mpoindexter: I have added the requested test.

The existing implementation did not handle query strings in the first place, so I added additional behavior to the `get_uri` method (since `get_url` already strips out query strings) that will preserve the query string.

w3af/plugins/crawl/open_api.py
mpoindexter: OK, removed this change and improved (hopefully!) the docs at the parameter level.

artem-smotrakov:
@andresriancho @mpoindexter like the idea of this patch!

mpoindexter:
@andresriancho - I think my latest changes address your comments.  Anything else or is this good to merge?

andresriancho:
Merged!

andresriancho:
giphy

Issue #17475, opened 2018-12-10
Error while reading plugin options: "No module named shell"


hktalent:
kali linux
```
cd ~/.py_virtualenv
virtualenv --no-site-packages w3af
source ~/.py_virtualenv/w3af/bin/activate
cd /mytools
git clone https://github.com/andresriancho/w3af
cd w3af
./w3af_console
/tmp/w3af_dependency_install.sh
./w3af_console


Error while reading plugin options: "No module named shell"

py2 ./w3af_console
/tmp/w3af_dependency_install.sh
py2 ./w3af_console

/usr/lib/python2.7/dist-packages/scipy/spatial/__init__.py:96: ImportWarning: Not importing directory '/usr/lib/python2.7/dist-packages/scipy/spatial/qhull': missing __init__.py
  from .qhull import *
/usr/lib/python2.7/dist-packages/scipy/optimize/_minimize.py:37: ImportWarning: Not importing directory '/usr/lib/python2.7/dist-packages/scipy/optimize/lbfgsb': missing __init__.py
  from .lbfgsb import _minimize_lbfgsb
It seems that your Python installation does not have all the modules required by the w3af framework. For more information about how to install and debug dependency issues please browse to http://docs.w3af.org/en/latest/install.html
Error while reading plugin options: "No module named shell"
```

hktalent:
```
# https://www.scipy.org/install.html
sudo apt-get install python-numpy python-scipy python-matplotlib ipython ipython-notebook python-pandas python-sympy python-nose
pip2 install  numpy scipy matplotlib ipython jupyter pandas sympy nose


#./w3af_console
/root/.py_virtualenv/w3af/local/lib/python2.7/site-packages/scipy/spatial/__init__.py:96: ImportWarning: Not importing directory '/root/.py_virtualenv/w3af/local/lib/python2.7/site-packages/scipy/spatial/qhull': missing __init__.py
  from .qhull import *
/root/.py_virtualenv/w3af/local/lib/python2.7/site-packages/scipy/optimize/_minimize.py:37: ImportWarning: Not importing directory '/root/.py_virtualenv/w3af/local/lib/python2.7/site-packages/scipy/optimize/lbfgsb': missing __init__.py
  from .lbfgsb import _minimize_lbfgsb
It seems that your Python installation does not have all the modules required by the w3af framework. For more information about how to install and debug dependency issues please browse to http://docs.w3af.org/en/latest/install.html
Error while reading plugin options: "No module named shell"
[...]

PR #17708, opened 2019-03-23
Fix a problem when using w3af on a case insensitive FS

mpoindexter:
When running w3af on a case-insensitive file system, the Halberd plugin has a problem where it tries to import itself vs. the real Halberd package, because they differ only by case.

I think this will fix https://github.com/andresriancho/w3af/issues/17475


andresriancho:
@mpoindexter where are you running w3af ?!

mpoindexter:
Hah! I'm doing development on a Mac, but using a Docker image with the code mounted to actually run the code since getting all the deps installed on OS X was a challenge.  So the OS is case sensitive, but the filesystem is not.

andresriancho:
:+1: thanks for the details

PR #17709, opened 2019-03-23
Fix errors reconfiguring open_api plugin

mpoindexter:
When setting configurations for open_api plugin in a profile, or
setting them multiple times from console, an error happens the
second time.

This happens because `set_value` is called with a `Headers` or
`QueryString` instance.  Then, in `validate` this is parsed as a
string, and an exception occurs.

### DIFF
--- a/w3af/core/data/options/header_option.py
+++ b/w3af/core/data/options/header_option.py
@@ -40,6 +40,8 @@ class HeaderOption(BaseOption):
         self._value = self.validate(value)
 
     def validate(self, value):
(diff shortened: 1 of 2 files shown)


andresriancho:
yes

andresriancho:
Just for the record, had to think about this one for a while... wasn't sure if the proposed solution was the best. It fixes the bug by "changing the interface", now those methods allow both strings and headers/qs instances to be used as params. Not 100% desirable, but acceptable :+1:

PR #17713, opened 2019-03-24
Fix a misleading display in the report for blindsqli

mpoindexter:
Without this change, delay based blind SQL injection
vulnerabilities are reported as occurring with the reversed
parameter, which doesn't make much sense.


andresriancho:
giphy-a

PR #17726, opened 2019-03-27
Fix an error when applying a profile then making config changes

mpoindexter:
When applying a profile, and then reconfiguring a plugin in the
console, the settings applied to the plugin could be lost. This
is caused by the fact that settings were not always applied to
the active plugin instance that would be used if it was
instantiated prior to the configuration operation.  This change
ensures that the latest settings are always applied to each
plugin.


andresriancho:
Thanks for the PR! The change looks good and I'll merge it, but in the future please provide a way for me to reproduce the issue, so I can better understand (besides just reading the code) what was fixed.

andresriancho:
giphy

PR #17733, opened 2019-03-28
Fix race condition

mpoindexter:
Occasionally inspect_threads() would raise an exeception.

This was caused by self.func becoming None after the check at the
top of get_real_func_name since inspect_threads() is called on a
different thread than self.func is set on.

Refactor the code such that checking the state of a thread never
mutates self.func/self.args, and there is no time of check/time of
use mismatch that can lead to an exception.


andresriancho:
amazing

PR #17734, opened 2019-03-29
Feature/fix race in inspect

mpoindexter:
Fix some bugs in special cases from my last PR to fix the race in inspecting the thread pool.  Sorry for this!


andresriancho:
Thanks!

PR #17738, opened 2019-03-30
fix docker scripts for GUI

DKingAlpha:


andresriancho:
Could you describe your PR? `python-webkit` dependency was missing, what is the rest?

DKingCN:
python-webkit is needed for GUI part running.

the container path is supposed to be `/home/w3af/.w3af` instead of `/root/.w3af`, so is `w3af-shard`. Profiles are stored in /home/w3af/.w3af instead of under root dir. idky but this is how its gonna work.

For the TypeError, when the env var is not set, None is returned, and int(None) throws TypeError instead of ValueError.

DOCKER_RUN_GUI, as you can see, is a startup cmdline for GUI.

PR #17739, opened 2019-03-31
fix mod_security that drops timeout argument

DKingAlpha:


andresriancho:
Thanks for your pull request, it looks good, but maybe we can take it one step further!

What if in the future we add one more attribute / parameter for `HTTPRequest`? We would need to change `mod_security` (and maybe many other code areas) again!

Could you find a way to do this in a way that will keep on working with future changes?

DKingCN:
```
def ex_call(func, mangle_dict, pass_obj):
    # python3: func.__code__.co_varnames
    paramlist = func.func_code.co_varnames  # python 2.7+
    kargs = {}
    for param in paramlist:
        kargs[param] = mangle_dict[param] if param in mangle_dict else pass_obj.__dict__[param]
    return func(**kargs)

# show case
def testx(a, b=1):
    print(a, b)

class ArgList:
    def __init__(self):
        self.a = 2
        self.b = 3

args = ArgList()
print(args.__dict__)
ex_call(testx, {'b': 5}, args)
```

actual usage: `ex_call(func=HTTPRequest, mangle_dict={...}, pass_obj=request)`

andresriancho:
Thanks for your proposed fix, but it was way too hackish! It was going to work... but was too complex to understand / maintain.

Based on your PR code / ideas / etc. I fixed all the evasion plugins in a way that will work over time.

Thanks!

PR #18128, opened 2019-09-10
added changes for fedora checks

404ram:
New fedora 30 packages for `python-pip` and `python-devel` renamed to `python2-*` and command for pip renamed to `pip`


andresriancho:
Thanks for the PR! The changes look good, but they are not backward-compatible. In other words, if I merge your change the dependency check for the old fedora distributions will fail.

Please take a look at how this is handled for Ubuntu (`ubuntu*.py` files in [this directory](https://github.com/andresriancho/w3af/tree/master/w3af/core/controllers/dependency_check/platforms)) and submit new changes that properly handle both old and new Fedora versions.

elpuffin:
Great thanks,

I will take a look when I get some free time and will let you know if I do
it.

Regards,

PR #18152, opened 2019-10-01
Feature/js

maczal:
I noticed the method name is currently prepended with underscore


andresriancho:
Thanks for finding that!

Fixed in a different way in https://github.com/andresriancho/w3af/commit/01a2271f3606730717c30ea486ab0c58264741ed

PR #18368, opened 2020-02-03
Feature/8561 cdn detection

### LINKED ISSUE #8561: CDN detection
Create a grep plugin that finds [these headers](https://github.com/WPO-Foundation/webpagetest/blob/master/agent/wpthook/cdn.h#L137) and reports an `Info` to the user telling him which CDN is being used to host the target web application.

Q-back:
Added CDN detection plugin, closes #8561


w3af/plugins/grep/cdn_providers.py
```diff
@@ -0,0 +1,255 @@
+from w3af.core.controllers.plugins.grep_plugin import GrepPlugin
```
andresriancho: Please include the copyright header.

w3af/plugins/grep/cdn_providers.py
```diff
@@ -0,0 +1,255 @@
...
+
+class cdn_providers(GrepPlugin):
+    """
+    Check CDN (Content Delivery Network) providers used by the website.
+
+    :author: Jakub Peczke (qback.contact@gmail.com)
+    """
+    # CDN headers stored in format ['header-name', 'header-value', 'provider's name']
+    cdn_headers = (
+        ["server", "cloudflare", "Cloudflare"],
```
andresriancho: All strings in w3af are single-quote delimited. PEP-8 says:

> In Python, single-quoted strings and double-quoted strings are the same. This PEP does not make a recommendation for this. Pick a rule and stick to it

So let's stick with single quotes :-)

w3af/plugins/grep/cdn_providers.py
```diff
@@ -0,0 +1,255 @@
...
+        [".llnwd.net", "Limelight"],
+        ["edgecastcdn.net", "Edgecast"],
+        [".systemcdn.net", "Edgecast"],
+        [".transactcdn.net", "Edgecast"],
+        [".v1cdn.net", "Edgecast"],
+        [".v2cdn.net", "Edgecast"],
+        [".v3cdn.net", "Edgecast"],
+        [".v4cdn.net", "Edgecast"],
+        [".v5cdn.net", "Edgecast"],
+        ["hwcdn.net", "Highwinds"],
```
andresriancho: Missing leading dot here? Or all the others don't need the leading dot?

w3af/plugins/grep/cdn_providers.py
```diff
@@ -0,0 +1,255 @@
...
+        [".cedexis.net", "Cedexis"],
+        [".streamprovider.net", "Rocket CDN"],
+    ]
+
+    def grep(self, request, response):
+        """
+        Check if responses are hosted by CDN provider
+        """
+        headers = response.get_headers()
+        for cdn_header in self.cdn_headers:
```
andresriancho: `for cdn_header in self.cdn_headers:` would look much better if you would use:

```python
for header_name, header_value, provider_name in self.cdn_headers:
```

That way it is easier to follow the code, and `cdn_header[1]` is replaced with the much better `header_value`

w3af/plugins/grep/cdn_providers.py
```diff
@@ -0,0 +1,255 @@
...
+            cdn_header_name = cdn_header[0]
+            if cdn_header_name in headers:
+                if cdn_header[1] == headers[cdn_header_name]:
+                    self._detect_cdn(response, cdn_header[2])
+
+        url = request.get_url()
```
andresriancho: Do you want to run URL detection even after the CDN was identified using the headers? This might yield duplicates in some cases?

Also, for a very very small % of requests, the ones that match to a CDN by domain; the `grep` method will be faster if you do the URL match first.

w3af/plugins/grep/cdn_providers.py
```diff
@@ -0,0 +1,255 @@
...
+        domain = url.get_domain()
+        for cdn_domain in self.cdn_domains:
+            if cdn_domain[0] in domain:
+                self._detect_cdn(response, cdn_domain[1])
+
+    def _detect_cdn(self, response, provider_name):
```
andresriancho: I would name this method `_save_cdn_finding_to_kb`

w3af/plugins/grep/cdn_providers.py
```diff
@@ -0,0 +1,255 @@
...
+            info,
+            group_klass=CDNProvidersInfoSet
+        )
+
+    def get_long_desc(self):
+        """
+        :return: A DETAILED description of the plugin functions and features.
+        """
+        return """
+        Check CDN (Content Delivery Network) providers used by the website.
```
andresriancho: `Identify any CDN (Content Delivery Network) providers used by the target website`

w3af/plugins/tests/grep/test_cdn_providers.py
```diff
@@ -0,0 +1,56 @@
+import unittest
```
andresriancho: Add missing copyright header

w3af/plugins/tests/grep/test_cdn_providers.py
```diff
@@ -0,0 +1,56 @@
...
+
+        # this request should be stored separately in kb as it comes from another provider
+        url = URL('https://example.com/cloudflare/')
+        request = FuzzableRequest(url, method='GET')
+        response = HTTPResponse(200, '', cloudflare_header, url, url, _id=3)
+        self.plugin.grep(request, response)
+
+        self.assertEqual(len(kb.get('cdn_providers', 'cdn_providers')), 2)
+
+    def test_if_cdn_provider_can_be_detected_by_url(self):
```
andresriancho: And add one more test that asserts that no finding is stored to the KB when a header that looks very much like (but is not) one of the CDNs is used.

andresriancho (changes requested):
Sorry for taking a lot of time to review this PR!

w3af/plugins/grep/cdn_providers.py
Q-back: I guess Highwinds is discovered only when we can't see any subdomain in url, while for example Netlify's url may look like `cdn.netlify.com`, so we get here `.netlify.com` to make subdomain is used.

w3af/plugins/grep/cdn_providers.py
Q-back: Good point, shame on me I didn't fix that before PR 😉

w3af/plugins/grep/cdn_providers.py
Q-back: You're right, I thought about case when plugin detects another CDN provider with header and another with URL. But in fact if someone adds wrong CDN info to header.. It just shouldn't happen. I changed it to return after detecting provider via header. URL is here as backup option if we can't detect CDN with header.

andresriancho:
giphy (1)

PR #18371, opened 2020-02-05
Feature/8992 improve backdoor signatures

### LINKED ISSUE #8992: Improve signatures to detect backdoors
## History
I refactored the find_backdoors plugin in order to use specific signatures, which might get outdated.

## Task
 - [ ] Find new signatures to detect web shells and backdoors
 - [ ] Update the signatures file using the shells at https://github.com/tennc/webshell
 - [ ] Update the signatures file using https://github.com/Neo23x0/Loki/blob/master/signatures/thor-webshells.yar , note that some signatures match source code and we don't have access to it, just to the rendered response
 - [ ] Create a task just like this one so I remember to keep doing this periodically

Q-back:
Grepped filenames and signatures from https://github.com/tennc/webshell. Closes #8992


w3af/plugins/crawl/find_backdoors/web_shells.txt
```diff
@@ -178,3 +178,721 @@ nc.exe
 netcat.exe
 socat.exe
 cmd.pl
+
```
andresriancho: Not 100% sure about this, but how are empty lines in this file handled by the plugin? Same with the previous file.

andresriancho:
Please confirm that the empty line is not an issue and we'll merge this!

w3af/plugins/crawl/find_backdoors/web_shells.txt
Q-back: Thanks to `line.strip()` empty lines are ignored. 
https://github.com/andresriancho/w3af/blob/193738bf46dd4d9476e6f632df4204ddf71227fb/w3af/plugins/crawl/find_backdoors.py#L103
```
if not line:
    continue
```
https://github.com/andresriancho/w3af/blob/193738bf46dd4d9476e6f632df4204ddf71227fb/w3af/plugins/crawl/find_backdoors.py#L108

andresriancho:
giphy

PR #18399, opened 2020-02-14
Fix crash when we try to free() terminated chrome instance in pool

Q-back:
Sometimes during scans the below exception was raised. It was probably caused by calling `pool.free()` on already terminated chrome instance.

```
A "AttributeError" exception was found while running crawl.web_spider on "Method: GET | https://domain/images/favicon.png". The exception was: "'NoneType' object has no attribute 'set_debugging_id'" at /usr/local/w3af/w3af/plugins/crawl/web_spider.py:_raise_chrome_crawler_exception()():633. The full traceback is:

  File "/usr/local/w3af/w3af/core/controllers/chrome/crawler/main.py", line 299, in _crawl_with_strategy_wrapper
    run_ro_strategies)
  File "/usr/local/w3af/w3af/core/controllers/chrome/crawler/main.py", line 338, in _crawl_with_strategy
    self._chrome_pool.free(chrome)
  File "/usr/local/w3af/w3af/core/controllers/chrome/pool.py", line 275, in free
    chrome.set_debugging_id(None)
  File "/usr/local/w3af/w3af/core/controllers/chrome/instrumented/main.py", line 94, in set_debugging_id
    super(InstrumentedChrome, self).set_debugging_id(debugging_id)
  File "/usr/local/w3af/w3af/core/controllers/chrome/instrumented/instrumented_base.py", line 188, in set_debugging_id
    self.chrome_conn.set_debugging_id(debugging_id)
AttributeError: 'NoneType' object has no attribute 'set_debugging_id'
```

### DIFF
--- a/w3af/core/controllers/chrome/instrumented/instrumented_base.py
+++ b/w3af/core/controllers/chrome/instrumented/instrumented_base.py
@@ -59,7 +59,6 @@ class InstrumentedChromeBase(object):
         self.http_traffic_queue = http_traffic_queue
 
         self.debugging_id = None
-        self.is_terminated = False
 
         self.proxy = self.start_proxy()
         self.chrome_process = self.start_chrome_process()
@@ -284,7 +283,6 @@ class InstrumentedChromeBase(object):
         self.proxy = None
         self.chrome_process = None
         self.chrome_conn = None
-        self.is_terminated = True
 
     def get_pid(self):
         try:
--- a/w3af/core/controllers/chrome/instrumented/instrumented_base.py
+++ b/w3af/core/controllers/chrome/instrumented/instrumented_base.py
@@ -59,6 +59,7 @@ class InstrumentedChromeBase(object):
         self.http_traffic_queue = http_traffic_queue
 
         self.debugging_id = None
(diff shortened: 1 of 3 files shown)


andresriancho:
This PR works, BUT it adds the minor complexity of having to keep `is_terminated` updated in order for everything to work well.

I would have fixed it with something similar to:

```python
if self.chrome_conn is None:
    return

self.chrome_conn.set_debugging_id(debugging_id)
```

Not sure where in the code that would be, but it makes more sense to me.

andresriancho:
Could you apply that change?

PS: In the future I'll be more relaxed with PR reviews, I know I'm being a PR-nazi now :-1:

Q-back:
I actually like PR-nazi style 😉 . I got used to it in my previous projects and I think it's a good approach because it helps to keep very high quality of code inside project. Imho it's worth to give it few extra minutes.

Q-back:
I use `chrome_process` instead of `chrome_conn` proposed by you @andresriancho because `chrome_process` name seems more verbose and it's cleaned up and initiated in the same place.

Issue #18430, opened 2020-03-02
Fix tests


Q-back:
# Tests fails in w3af

Currently I'm not able to just run nosetest and make it pass. At some point tests hangs and I kill it after 10 minutes of inactivity. I was able to accomplish test run by skipping some performance tests. The final result was:
```
Ran 2583 tests in 1587.648s
FAILED (SKIP=33, errors=886, failures=131)
```
So generally we have about 2.6k tests, it's impressive, but as I can see there are really many tests which should be refactored or shouldn't run under normal conditions. I mean all performance tests, all tests which depend on another tests or test which communicate with outside world.

## Solution
My idea is to write script which will:
* mark all tests which fails as deprecated and skip them
* mark all tests which takes longer than few seconds as performance and skip them

Right now we can’t run passing test suite anyway, so it’s better than nothing. After that when developer works on feature x and see some tests skipped he can look after them and fix them.

I'd like to switch to pytest. Nose seems to be a bit abandoned as last commit is from 2016 https://github.com/nose-devs/nose. Also pytest dependency injection is really powerful.
In general I think all tests should be passing under 5 minutes. Then we're good.

andresriancho:
👏

Q-back:
Done https://github.com/andresriancho/w3af/commit/02aed8190cb8872ad9502d94e4efed5535b8ee7f

Q-back:
Done https://github.com/andresriancho/w3af/commit/02aed8190cb8872ad9502d94e4efed5535b8ee7f

PR #18431, opened 2020-03-02
Fix/recursive crawl

Q-back:
There are Vue/React/Angular pages which instead of showing 404 error message redirect us to the main page (like `this.$route.replace('/')`). When we scan the page like this JS crawler goes for infinite loop and discovers deeply nested URLs forever with the same response.

Here's simple fix which can potentially prevent this bug for most cases.
We can't use `VariantDB.MAX_PATH_VARIANTS` as it would affect many crawling aspects and limit crawling flow. We want to limit only *recursively nested path crawling*, not the *path variants crawling*.

## TODO
- [ ] ensure `_request_is_looped` in VariantDB won't stop whole scan too early. What if we'll detect looped URL before crawling the rest of URLs?
- [ ] test if it won't destroy proper crawling (false positives)
- [x] what if the main page includes e.g. current timestamp? Response hash could be different

Also I wasn't able to write an integration test for this issue, as I wasn't able to simulate the server URL resolving conditions. Blocked by https://github.com/andresriancho/w3af/issues/18430.
But finally I was able to at least write unit test, so hopefully it can be enough for now.

### DIFF
--- a/w3af/core/data/db/variant_db.py
+++ b/w3af/core/data/db/variant_db.py
@@ -19,6 +19,7 @@ along with w3af; if not, write to the Free Software
 Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 """
+import os
 import threading
 
 import w3af.core.data.kb.config as cf
@@ -48,6 +49,8 @@ from w3af.core.data.db.clean_dc import (clean_fuzzable_request,
 # paths inside the "abc" path.
 #
 PATH_MAX_VARIANTS = 50
+
+PATH_MAX_LOOP = 4
 
 #
 # Limits the max number of variants we'll allow for URLs with the same path
@@ -118,6 +121,7 @@ class VariantDB(object):
 
         self.params_max_variants = cf.cf.get('params_max_variants')
         self.path_max_variants = cf.cf.get('path_max_variants')
+        self.path_max_loop = cf.cf.get('path_max_loop')
         self.max_equal_form_variants = cf.cf.get('max_equal_form_variants')
 
         self._db_lock = threading.RLock()
@@ -137,6 +141,10 @@ class VariantDB(object):
         with self._db_lock:
             if self._seen_exactly_the_same(fuzzable_request, request_hash):
                 self._log_return_false(fuzzable_request, 'seen_exactly_the_same')
+                return False
+
+            if self._request_is_looped(fuzzable_request):
+                self._log_return_false(fuzzable_request, 'request_is_looped')
                 return False
 
             if self._has_form(fuzzable_request):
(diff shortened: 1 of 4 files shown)


Q-back:
As response hash can be different for every page including e.g. timestamp and also response hash may be exactly the same for every page when react/vue/anguar app is not using server side rendering we may also move from hashing responses to sophisticated URL checking.

Q-back:
Moved to URL checking in `VariantDB`.

w3af/core/data/db/variant_db.py
```diff
@@ -201,6 +209,32 @@ def _seen_exactly_the_same(self, fuzzable_request, request_hash):
...
+        same_path_parts = 0
+        # first iteration will almost never increment same_path_parts, because
+        # first head will be probably something like 'index.html'
+        for _ in range(self.path_max_loop + 1):
+            prev_head = head
+            path, head = os.path.split(path)
+            if head and head == prev_head:
+                same_path_parts += 1
+
+        if same_path_parts >= self.path_max_loop:
```
andresriancho: I freaking love your solution. It is amazingly simple!

w3af/plugins/crawl/web_spider.py
```diff
@@ -19,6 +19,7 @@
 Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 """
+import hashlib
```
andresriancho: Remove

w3af/plugins/crawl/web_spider.py
```diff
@@ -72,6 +73,7 @@ def __init__(self):
         self._target_domain = None
         self._already_filled_form = ScalableBloomFilter()
         self._variant_db = VariantDB()
+        self._found_responses = {}
```
andresriancho: Remove

w3af/plugins/crawl/web_spider.py
```diff
@@ -432,11 +434,56 @@ def _extract_links_and_verify(self, doc_parser, fuzzable_req, http_response, deb
...
+        over and over again but under another url.
+
+        We want to store hash of response body, so if the same hash will occur again
+        we'll know the response isn't unique. Calculating hash is CPU-expensive,
+        so we store only length of response in self._found_responses dict.
+        If we find second response with same length then we calculate hash for it
+        and store it like
+        {
+            123: [],  # no hash, the resp with length 123 was found only once
+            24: ['dsf42434lkaf9j@3ls', 'd3#-dfr3'],  # two different responses with
```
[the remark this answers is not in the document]
Q-back: I've decided to not implement the fix this way. Should I bring it back?

w3af/plugins/crawl/web_spider.py
```diff
@@ -432,11 +434,56 @@ def _extract_links_and_verify(self, doc_parser, fuzzable_req, http_response, deb
...
+        :returns bool: return True if response body is brand new and unique and False if
+        we have already crawled the same response body.
+        """
+        # We want to first store only length of response as it's the fastest way
+        # to check if they're quite unique
+        body_length = len(response_body)
+
+        # Check if response with same length was already crawled
+        similar_response_hashes = self._found_responses.get(body_length)
+        if similar_response_hashes:
```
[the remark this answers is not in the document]
Q-back: I've decided to not implement the fix this way. Should I bring it back?

w3af/plugins/crawl/web_spider.py
```diff
@@ -432,11 +434,56 @@ def _extract_links_and_verify(self, doc_parser, fuzzable_req, http_response, deb
                                                                  fuzzable_req,
                                                                  http_response,
                                                                  debugging_id)
+        is_response_unique = self._ensure_response_is_unique(http_response.body)
```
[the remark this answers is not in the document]
Q-back: I've decided to not implement the fix this way. Should I bring it back?

w3af/core/data/db/variant_db.py
Q-back: I'm only afraid that returning `False` in `VariantDB.append()` will finish whole scan before web_spider will crawl the rest of URLs. Do you know if it will work this way?

andresriancho: The scan will not stop if you return false in `append()`. The only action that should be done after `False` is returned is to ignore that "looped URL".

PR #18492, opened 2020-04-06
Feature/vulndb

Q-back:
Added `vulndb_id` according to specification from Holm Security. Also some `Info` instances were turned into `Vuln` instances with `INFORMATION` severity. My concept is to make `vulndb_id` param required in `Vuln.__init__` while `Info` wouldn't require `vulndb_id`.

### DIFF
--- a/w3af/core/controllers/plugins/auth_plugin.py
+++ b/w3af/core/controllers/plugins/auth_plugin.py
@@ -26,8 +26,9 @@ import w3af.core.data.kb.knowledge_base as kb
 import w3af.core.data.kb.config as cf
 
 from w3af.core.controllers.plugins.plugin import Plugin
-from w3af.core.data.kb.info import Info
+from w3af.core.data.constants import severity
 from w3af.core.data.fuzzer.utils import rand_alnum
+from w3af.core.data.kb.vuln import Vuln
 from w3af.core.data.url.helpers import is_no_content_response
 
 
@@ -278,10 +279,12 @@ class AuthPlugin(Plugin):
                        '%s')
             desc = msg_fmt % args
 
-        i = Info(title,
+        i = Vuln(title,
                  desc,
+                 severity.INFORMATION,
                  self._http_response_ids,
-                 self.get_name())
+                 self.get_name(),
+                 vulndb_id=10116)
 
         i.set_uri(self._get_main_authentication_url())
 
--- a/w3af/core/controllers/sql_tools/blind_sqli_time_delay.py
+++ b/w3af/core/controllers/sql_tools/blind_sqli_time_delay.py
@@ -130,7 +130,7 @@ class BlindSQLTimeDelay(object):
 
             v = Vuln.from_mutant('Blind SQL injection vulnerability', desc,
                                  severity.HIGH, response_ids, 'blind_sqli',
-                                 mutant)
+                                 mutant, vulndb_id=46)
 
             om.out.debug(v.get_desc())
 
--- a/w3af/plugins/audit/cors_origin.py
+++ b/w3af/plugins/audit/cors_origin.py
@@ -167,7 +167,14 @@ class cors_origin(AuditPlugin):
             msg %= (url, ACCESS_CONTROL_ALLOW_METHODS,
                     allow_methods, ', '.join(report_sensitive))
 
-            v = Vuln(name, msg, severity.LOW, response.get_id(), self.get_name())
+            v = Vuln(
+                name,
+                msg,
+                severity.LOW,
+                response.get_id(),
+                self.get_name(),
+                vulndb_id=10010,
+            )
             v.set_url(forged_req.get_url())
             v[METHODS] = allow_methods_set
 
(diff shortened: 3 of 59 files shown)


w3af/core/controllers/plugins/auth_plugin.py
```diff
@@ -278,10 +279,12 @@ def _log_info_to_kb(self, title, message, include_log_messages=True):
...
             desc = msg_fmt % args
 
-        i = Info(title,
+        i = Vuln(title,
                  desc,
+                 severity.INFORMATION,
                  self._http_response_ids,
-                 self.get_name())
+                 self.get_name(),
+                 vulndb_id=10116)
```
andresriancho, line 287: https://github.com/vulndb/data/search?q=10116&unscoped_q=10116

What happens when the ID is not present in the `vulndb`? Is this error handled somewhere?

w3af/core/controllers/sql_tools/blind_sqli_time_delay.py
```diff
@@ -130,7 +130,7 @@ def is_injectable(self, mutant, delay_obj):
 
             v = Vuln.from_mutant('Blind SQL injection vulnerability', desc,
                                  severity.HIGH, response_ids, 'blind_sqli',
-                                 mutant)
+                                 mutant, vulndb_id=46)
```
andresriancho, line 133: I believe there is not need to specify the `vulndb_id` parameter here. If the title specified in the `Vuln.from_mutant()`: `Blind SQL injection vulnerability` matches the title in the vulndb, then the information is associated automatically. That wasn't working? Why adding the explicit ID if it was?

w3af/plugins/audit/cors_origin.py
```diff
@@ -167,7 +167,14 @@ def _allow_methods(self, forged_req, url, origin, response,
...
                     allow_methods, ', '.join(report_sensitive))
 
-            v = Vuln(name, msg, severity.LOW, response.get_id(), self.get_name())
+            v = Vuln(
+                name,
+                msg,
+                severity.LOW,
+                response.get_id(),
+                self.get_name(),
+                vulndb_id=10010,
```
andresriancho, line 176: Would it be possible for Holm to contribute back to the `vulndb`? If we do that, all the w3af users would get these detailed descriptions for the vulnerabilities.

w3af/plugins/output/xml_file/finding.tpl
```diff
@@ -1,4 +1,4 @@
-<vulnerability id="{{ id_list }}" method="{{ http_method | escape_attr }}" name="{{ name | escape_attr }}" plugin="{{ plugin_name }}" severity="{{ severity }}" url="{{ url | escape_attr }}" var="{{ var | escape_attr }}">
+<vulnerability id="{{ id_list }}" method="{{ http_method | escape_attr }}" name="{{ name | escape_attr }}" plugin="{{ plugin_name }}" severity="{{ severity }}" url="{{ url | escape_attr }}" var="{{ var | escape_attr }}" vulndb_id="{{ vulndb_id }}">
```
andresriancho, line 1: Maybe `vulndb_id={{ vulndb_id }}` instead of `vulndb_id="{{ vulndb_id }}"`? Those IDs are ints. (just a detail)

andresriancho (changes requested):

w3af/core/controllers/plugins/auth_plugin.py
Q-back: There's no error. I don't implement here the concept of making `vulndb_id` required for `Vuln` instances. I just mentioned the concept in description, because in this PR I prepare the code for this concept. I'll be able to implement it once https://github.com/andresriancho/w3af/pull/18543 is merged.

w3af/core/controllers/sql_tools/blind_sqli_time_delay.py
Q-back: Yes, it was working. However it won't currently work for vulns with ID 10000 and higher. I'd opt to leave it as it is because it doesn't change anything at all for lower IDs and for higher IDs it allows to assign vulndb_id.

w3af/plugins/audit/cors_origin.py
Q-back: I brought up the discussion internally in Holm.

w3af/core/controllers/plugins/auth_plugin.py
andresriancho: My point is, when `10116` is passed to the `Vuln` instance, and then XML plugin tries to write that vuln instance to disk, what happens? Will it crash because `10116` is NOT in the `vulndb`?

`10116` is Holm-specific, that is why I'm asking.

w3af/plugins/audit/cors_origin.py
andresriancho: And what was the resolution?

Q-back: So far there's no decision to contribute to `vulndb`. Holm holds off

PR #18543, opened 2020-05-08
Fix/tests

Q-back:
I wrote a script which automatically marked all deprecated and slow tests. You can check how it was done in the commits under this branch: https://github.com/Q-back/w3af/commits/fix/tests-after-script. Now it's as simple as run `pytest` in w3af root directory, and you'll get tests passed with no errors under 5 minutes 🎉 ! More than 1000 tests were marked as slow or deprecated.
Pytest output:
`1525 passed, 53 skipped, 1025 deselected, 2 warnings in 248.85 seconds`
Please merge it ASAP and I'll go further to bring tests back into w3af contribution workflow 🚀 .
closes https://github.com/andresriancho/w3af/issues/18430


pytest.ini
```diff
@@ -0,0 +1,5 @@
+[pytest]
```
andresriancho, line 1: Do we need to add this file and `conftest.py` to the project root? Is there any other way to do it in order to avoid polluting the project root?

w3af/core/controllers/daemons/tests/test_webserver.py
```diff
@@ -19,6 +19,7 @@
 Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 """
+import pytest
```
andresriancho, line 22: In some cases you're adding `import pytest` and you really don't need it.

Maybe modify the script you wrote to avoid this?

andresriancho (changes requested):
Minor comments.

Would also be nice to have a modification to the project documentation (both wiki and read the docs markdown) on how to run unittests now!

w3af/core/controllers/daemons/tests/test_webserver.py
Q-back: script was modified few times between runs and right now it's not a one-time run script, but I was using it incrementally due to some problems. I would like to not run it again ;). But I can check for unused pytest imports

pytest.ini
Q-back: I've deleted `conftest.py`. However I opt for leaving `pytest.ini` in project root. 
* The `pytest.ini` file allows pytest to indentify project root. 
* It's good clue for developer that we're using `pytest` for running tests.
* It modifies how `pytest` command works, especially in our case it may be not obvious (because it's configured to skip many tests), so it's good to have it in visible place
* Other big projects also keep `pytest.ini` in root dir, for example the [python requests](https://github.com/psf/requests) library

I have never seen a project which would keep `pytest.ini` in different place than project root, I guess it may be also tricky to show pytest proper root_dir in such case.
As I can see we also have `circle.yml` in the project root, so `pytest.ini` is the same kind of file.

Q-back:
@andresriancho Please re-check conversations. I've also added tests to docs, I didn't do that yet in project wiki. I'm going to write about it after tests will get merged to master 😉 .

pytest.ini
Q-back: An alternative could be also `tox.ini` file.

andresriancho:
LGTM but now the PR has conflicts.

Feel free to resolve all this and commit+push directly to develop.

PR #18552, opened 2020-05-13
Feature/improve autocomplete js

Q-back:
- autocomplete_js is able to detect more inputs/buttons with common CSS selectors (check `dom_analyzer.js`)
- autocomplete_js runs `has_active_session` method using the same chrome instance as `_do_login()` method to preserve already logged in instance
- added new options to autocomplete_js config, so user can provide his own CSS selectors to find login button and username input. It should cover edge cases when we're not able to automatically detect them:
  * `username_field_css_selector`
  * `login_button_css_selector`
- added new option to autocomplete_js config, so user can provide CSS selector for button that needs to be clicked to display login form:
  * `login_form_activator_css_selector`

### DIFF
--- a/w3af/core/controllers/chrome/instrumented/main.py
+++ b/w3af/core/controllers/chrome/instrumented/main.py
@@ -23,6 +23,7 @@ import time
 import json
 
 import w3af.core.controllers.output_manager as om
+from w3af.core.controllers.chrome.devtools.exceptions import ChromeScriptRuntimeException
 
 from w3af.core.data.parsers.doc.url import URL
 from w3af.core.controllers.chrome.instrumented.instrumented_base import InstrumentedChromeBase
@@ -307,8 +308,8 @@ class InstrumentedChrome(InstrumentedChromeBase):
             'window._DOMAnalyzer.getLoginForms("{}", "{}")'
         )
         func = func.format(
-            exact_css_selectors.get('username_input', ''),
-            exact_css_selectors.get('login_button', ''),
+            exact_css_selectors.get('username_input', '').replace('"', '\\"'),
+            exact_css_selectors.get('login_button', '').replace('"', '\\"'),
         )
         result = self.js_runtime_evaluate(func)
 
@@ -335,8 +336,8 @@ class InstrumentedChrome(InstrumentedChromeBase):
             'window._DOMAnalyzer.getLoginFormsWithoutFormTags("{}", "{}")'
         )
         func = func.format(
-            exact_css_selectors.get('username_input', ''),
-            exact_css_selectors.get('login_button', ''),
+            exact_css_selectors.get('username_input', '').replace('"', '\\"'),
+            exact_css_selectors.get('login_button', '').replace('"', '\\"'),
         )
         result = self.js_runtime_evaluate(func)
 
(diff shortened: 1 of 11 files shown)


w3af/core/controllers/chrome/js/dom_analyzer.js
```diff
@@ -865,6 +865,54 @@ var _DOMAnalyzer = _DOMAnalyzer || {
...
+        result = document.querySelectorAll("input[type='submit']", parentElement);
+        if (!result.length) {
+            result = document.querySelectorAll("button[type='submit']", parentElement);
+        }
+        // Maybe it's just normal button with innerText: 'Login'...
+        if (!result.length) {
+            result = [];
+            let buttons = document.querySelectorAll('button', parentElement);
+            for (let button of buttons) {
+                if (button.innerText.toLocaleLowerCase().includes('log')) {
```
andresriancho, line 908: I usually dislike code that is tied to a specific language. `log` might work for `login` and `logga` (which I believe is login in `.se`) but... for sites developed in other languages (spanish, polish, etc.) this will not work.

Also, `sign-in` will not be found.

Can we find an alternative?

w3af/plugins/auth/autocomplete_js.py
```diff
@@ -256,22 +273,71 @@ def _find_form_submit_strategy(self, chrome, form):
...
+        :returns OptionList: list of option objects for plugin
+        """
+        option_list = super(autocomplete_js, self).get_options()
+        autocomplete_js_options = [
+            (
+                'username_field_css_selector',
+                self.username_field_css_selector,
+                STRING,
+                "(Optional) Exact CSS selector which will be used to retrieve "
+                "the username input field. If provided then w3af won't try "
```
andresriancho, line 309: In most cases I was trying NOT to mention `w3af` in things that the end-user might see. If you guys ever expose this to the user, they will know that w3af is used in the backend.

Maybe change to: `When provided the scanner is not going to try to detect the input field in an automated way`

Same problem below.

w3af/plugins/auth/autocomplete_js.py
```diff
@@ -256,22 +273,71 @@ def _find_form_submit_strategy(self, chrome, form):
...
+    def set_options(self, options_list):
+        options_list_copy = deepcopy(options_list)  # we don't want to touch real option_list
+        self.username_field_css_selector = options_list_copy.pop(
+            'username_field_css_selector'
+        ).get_value()
+        self.login_button_css_selector = options_list_copy.pop(
+            'login_button_css_selector'
+        ).get_value()
+        super(autocomplete_js, self).set_options(options_list_copy)
+
```
andresriancho, line 340: Add the new configuration parameters to `get_long_desc`

Add some documentation explaining how they work, when to use them, etc.

andresriancho (changes requested):
The code is really good! 💯 

Just minor changes and comments.

w3af/core/controllers/chrome/js/dom_analyzer.js
Q-back: The only thing that comes into my mind is to return all buttons located inside `parentNode`. Hopefully there will be login button.

andresriancho:
> This pull request is still a work in progress

@Q-back is this ready for a second review?

Q-back:
@andresriancho Thanks for activity 😉 . There's one new improvement which I'd like to finish before review.

w3af/core/controllers/chrome/instrumented/main.py
```diff
@@ -589,19 +608,13 @@ def js_runtime_evaluate(self, expression, timeout=5):
...
-        if 'result' not in result['result']:
-            return None
-
-        if 'value' not in result['result']['result']:
-            return None
-
-        return result['result']['result']['value']
+        runtime_exception = result.get('result', {}).get('exceptionDetails')
+        if runtime_exception:
+            raise ChromeScriptRuntimeException(
```
andresriancho, line 613: This is new! Are you sure it is handled in all calls to this method? In the past we were returning `None` and now we raise an exception. Make sure to search all calls for this method and modify error handling appropiately.

Q-back: Yeah, I have added few more fixes in this PR.
This works for sure. It's additional (new) exception which is raised when underlying script throws internal JS exception. We still return `None` in other cases here https://github.com/andresriancho/w3af/pull/18552/files/8d520bc067455089a9fccac9c395f9114f521e8a#diff-4a954d4e5c6123fad237c66f5399b9dfR617

PR #18619, opened 2020-06-16
added github test action

Q-back:
blocked by https://github.com/andresriancho/w3af/issues/18622 as we won't be able to run any tests which uses chrome/chromium. So right now tests are failing


Q-back:
Seems like GitHub actions makes w3af able to detect chrome/chromium callable, but it's updated to latest version, so w3af fails running it. #18622 is still blocker in this case

PR #18630, opened 2020-06-26
autocomplete_js improvements

Q-back:
1. It includes merge from https://github.com/andresriancho/w3af/pull/18612
2. It includes merge from https://github.com/andresriancho/w3af/pull/18552
3. It adds info about failed CSS selectors to report file.
4. Added `TestPluginRunner` class which is a tool for testing every plugin with clean API. It's intended to be a simple to use fixture.

PR #18645, opened 2020-07-14
contribution at 14-07-2020

Q-back:
- includes all changes from https://github.com/andresriancho/w3af/pull/18630
- adds SOAP plugin


w3af/plugins/crawl/soap.py
```diff
@@ -0,0 +1,168 @@
+import contextlib
+import sys
+from cStringIO import StringIO
+
+import zeep
```
andresriancho: Did you add this to the requirements file(s)?

w3af/plugins/crawl/soap.py
```diff
@@ -0,0 +1,168 @@
...
+        super(soap, self).__init__()
+        self._is_first_run = True
+        self.service_url = ''
+        self.verbosity_level = 0
+
+    def crawl(self, fuzzable_request, debugging_id):
+        url_to_crawl = fuzzable_request.get_url().url_string
+        if self._is_first_run:
+            self._is_first_run = False
+            url_to_crawl = self.service_url
```
andresriancho: On the first call you lose the `fuzzable_request` URL. Instead of doing that, maybe you can perform the analysis twice, once on the `self.service_url` and once on `fuzzable_request`.

It is very unlikely, but maybe the `fuzzable_request` contains valuable data?

w3af/plugins/crawl/soap.py
```diff
@@ -0,0 +1,168 @@
...
+    def get_options(self):
+        option_list = super(soap, self).get_options()
+
+        option = opt_factory(
+            'service_url',
+            default_value=self.service_url,
+            desc='URL pointing to WSDL specification',
+            help=(
+                "Web applications which are based on SOAP protocol exposes their"
+                "specification in WSDL format. The WSDL specification gives us info"
```
andresriancho: There are some missing spaces here. For example `theirspecification` will be displayed.

w3af/plugins/crawl/soap.py
```diff
@@ -0,0 +1,168 @@
...
+            default_value=self.verbosity_level,
+            desc='Verbosity level',
+            help=(
+                "You can specify number as below:\n"
+                "0: No verbosity. Plugin will only detect new URLs.\n"
+                "1: Plugin will detect URLs and it will add info to knowledge base"
+                "about possible operations on given URL.\n"
+                "2: Plugin will detect URLs and dump all info from WSDL spec."
+            ),
+            _type=option_types.INT,
```
andresriancho: I believe there is an option type like "dropdown" or similar where you could specify the options (0, 1, 2) and it will validate that the selected option is in the provided range.

w3af/plugins/crawl/soap.py
```diff
@@ -0,0 +1,168 @@
...
+
+    def set_options(self, options_list):
+        super(soap, self).set_options(options_list)
+        self.service_url = options_list['service_url'].get_value()
+        self.verbosity_level = options_list['verbosity_level'].get_value()
+
+    def get_long_desc(self):
+        return (
+            "This plugin parses SOAP according to provided WSDL specification."
+            "User should specify the WSDL URL."
```
andresriancho: Add better, longer, description here.

w3af/plugins/crawl/soap.py
```diff
@@ -0,0 +1,168 @@
...
+            self._report_wsdl_dump(dump_capturer.getvalue())
+
+        for service_name in wsdl_client.wsdl.services:
+            service = wsdl_client.bind(service_name)
+            ports = wsdl_client.wsdl.services[service_name].ports
+            for _, port in ports.items():
+                operation_url = port.binding_options['address']
+                if self.verbosity_level == 1 and operation_url not in discovered_urls:
+                    self._report_all_operations_for_url(service, operation_url)
+                discovered_urls.add(operation_url)
```
andresriancho: What about HTTP request parameters? I don't see any handling of POST-body nor query string parameters.

andresriancho: I recommend moving the parsing of WSDL to the `parsers` path, where all the other parsers live. The WSDL parser will eventually grow and we don't want it here in the plugin.

I'm not 100% sure that we need this plugin. If you add the WSDL parser to:
https://github.com/andresriancho/w3af/blob/master/w3af/core/data/parsers/document_parser.py#L37

And w3af finds the WSDL URL, it will be able to parse it in `web_spider` and return Fuzzable HTTP requests, which can then be translated into a WSDL mutant, similar to:

https://github.com/andresriancho/w3af/blob/master/w3af/core/data/fuzzer/mutants/json_mutant.py
https://github.com/andresriancho/w3af/blob/master/w3af/core/data/fuzzer/mutants/xmlrpc_mutant.py (this one doesn't work)

The user would just have to add the WSDL to the target URL(s) of w3af, without even having to enable a plugin and set the target there. Also, another benefit of this approach, would be that the user doesn't even know that there is a WSDL endpoint, w3af finds it and analyzes it.

w3af/plugins/crawl/soap.py
Q-back: fixed

w3af/plugins/crawl/soap.py
Q-back: Not longer an issue, SOAP plugin is refactored to `wsdl_parser`

w3af/plugins/crawl/soap.py
Q-back: Not longer an issue, SOAP plugin is refactored to `wsdl_parser`

w3af/plugins/crawl/soap.py
Q-back: Not longer an issue, SOAP plugin is refactored to `wsdl_parser`

w3af/plugins/crawl/soap.py
Q-back: Not longer an issue, SOAP plugin is refactored to `wsdl_parser`

w3af/plugins/crawl/soap.py
Q-back: `wsdl_parser` created.

w3af/core/controllers/chrome/login/submit_form/main.py
```diff
@@ -91,3 +93,4 @@ def _handle_exception(self, strategy, e):
                 e,
                 self.debugging_id)
         om.out.debug(msg % args)
+        om.out.error(traceback.format_exc())
```
andresriancho: This was just added for debugging?

w3af/core/data/parsers/doc/wsdl.py
```diff
@@ -19,194 +19,110 @@
...
+
+    def __setstate__(self, state):
+        self.__dict__.update(state)
+        self._wsdl_client = zeep.Client(str(self._http_response.get_uri()))
+
+    @staticmethod
+    def can_parse(http_resp):
+        url = http_resp.get_uri()
+        try:
+            wsdl_client = zeep.Client(str(url))
```
andresriancho: This will send an HTTP request to the URL *without* using the w3af HTTP client. This means that, for example, HTTP proxy configurations are not going to be respected. For some cases that will break the feature. Is there a way to create a wsdl_client by sending the WSDL URL and HTTP response content?

w3af/core/data/parsers/doc/wsdl.py
```diff
@@ -19,194 +19,110 @@
...
+            name='SOAP details',
+            desc=dump_string,
+            response_ids=[],
+            plugin_name='web_spider',
+        )
+        kb.kb.append(self, 'soap_actions', dump_info)
+
+    @staticmethod
+    def get_name():
+        return 'wsdl_parser'
```
andresriancho, line 197: Maybe I read too quick, but where is the parser returning the requests? You should have something similar to:

https://github.com/andresriancho/w3af/blob/master/w3af/core/data/parsers/doc/open_api/main.py#L201
https://github.com/andresriancho/w3af/blob/master/w3af/core/data/parsers/doc/open_api/main.py#L250

andresriancho (changes requested):

w3af/core/controllers/chrome/login/submit_form/main.py
Q-back: It was added by purpose as I think it's good to have at least traceback saved when exception is silenced. If I wouldn't add this there would be no way to find this traceback as exception is silenced at this point. Showing only the exception name is not enough in many cases. How should be handled this one then?

w3af/core/data/parsers/doc/wsdl.py
Q-back: While `OpenAPI().get_api_calls()` is a special method used only by `open_api` crawl plugin I've resigned from SOAP plugin according to [this comment](https://github.com/andresriancho/w3af/pull/18645#discussion_r454315477). So `wsdl` parser only provides the standard `get_references` method which is inherited from `BaseParser` and used by `web_spider`.  `get_references` method provides results as `URL` instances not `FuzzableRequest` instances. Anyway I guess that in my current flow `WSDLMutant` is missing.

Please let me know if above is the right approach.

w3af/core/controllers/chrome/login/submit_form/main.py
andresriancho: I'm not sure about the context here, and if it will work, but we have an exception handler that will catch "all exceptions" and save them to a text file in /tmp.

Another alternative, if the scan can continue without a major reduction in functionality and user doesn't need to know about this issue is to also use `debug`?

w3af/core/data/parsers/doc/wsdl.py
andresriancho: If your parser only returns URLs, then unless there is a SOAP endpoint which takes query string parameters... you won't be able to find any vulnerabilities based on the output of this parser.

I strongly recommend you add some code that will allow you to retrieve fuzzable requests (or was it mutants? I forgot) from the parser.

andresriancho: Maybe you might need a custom plugin after all? OpenAPI document parser has its own plugin to extract the HTTP requests

https://github.com/andresriancho/w3af/blob/feature/js/w3af/plugins/crawl/open_api.py
https://github.com/andresriancho/w3af/blob/feature/js/w3af/plugins/crawl/open_api.py#L217

Now that we might have two parsers (openapi, wsdl) with their own plugins (eeek), maybe it is time to do an analysis and understand if we need to generalize all this into `web_spider`? Not sure... but worth thinking about it

w3af/core/data/parsers/doc/wsdl.py
Q-back: fix in https://github.com/andresriancho/w3af/pull/18645/commits/a81845fe4687d352b2e641da52bd6564632a320e

w3af/core/data/parsers/doc/wsdl.py
```diff
@@ -21,23 +21,62 @@
...
+    As WSDLParser has to be tight coupled to Zeep by design we have to also
+    make tight coupling between WSDLParser and ExtendedUrllib. And that's because
+    parser by design is not intended to perform any requests by itself. Although
+    Zeep is constructed in this specific way that it performs request when it's
+    instantiated.
+    As parsers are not intended to make requests there's also no obvious way to
+    pass uri_opener into parser.
+    """
+    def __init__(self):
+        super(ZeepTransport, self).__init__(timeout=2)
```
Q-back: leftover after debugging, fixed in later commit.

w3af/core/controllers/chrome/login/submit_form/main.py
Q-back: > we have an exception handler that will catch "all exceptions" and save them to a text file in /tmp

Not this time. The exception is not propagated in this place, it's simply silenced.
Changed to `debug` according to your recommendation c3e0eec2b.

w3af/core/data/parsers/doc/wsdl.py
Q-back: Fixed in https://github.com/andresriancho/w3af/pull/18645/commits/1a063e2341f2efc439f5f38107ba35ad577e2a28. I've added method `get_fuzzable_requests` to `BaseParser` as `WSDLParser` is second parser (after `OpenAPI`) which will produce fuzzable requests. `BaseParser.get_fuzzable_requests` returns empty list by default instead of raising `NotImplementedError`, as I don't see any point of raising exception in base class (which is a kind of interface or abstract probably) and overriding it in every subclass to return empty list. It's better to provide handy default in base class already. 
Also in future we may also want to refactor `OpenAPI` to implement `get_fuzzable_requests` and resign from openapi plugin, but it will break compatibility with current release by deleting one of plugins, so I opt for doing it later, ideally after moving to semantic versioning system.

